Issue #453, opened 2016-05-15
avalara tax module does not work ==> taxevaluation context is not set correctly


alexdoan102:
avalara tax module does not work ==> taxevaluation context is not set correctly
 ==> the taxevaluationcontext.Address & taxevaluationcontext.Customer  are BOTH NULL.
       ==> since they are both NULL the avalara service is not called at all.

![c1_taxevalcontext_avalara](https://cloud.githubusercontent.com/assets/73307/15271552/e117bdd4-1a1d-11e6-9061-87ed3759a112.png)

aartiomov:
Hi, I'm developer of the Avalara module. It seems that TaxEvaluationContext has changed recently. Let us check and look for the solution asap.
Thanks.

Woland2k:
@aartiomov can we make sure to include some xunit tests for avalara so such error can be cought during automated tests?

alexdoan102:
@tatarincev @aartiomov , I created PULL Request #479 . This fixes that the address and customerid so that the avalaraTax module would execute.
However, the avalara service calls returns "At least one address or LocationCode is required. Address is incomplete or invalid.". I verified that I am sending the correct address. (see screenshot). Please advise.
![avalarabug](https://cloud.githubusercontent.com/assets/73307/15662316/6b6ea44c-26bd-11e6-96c8-d0227d967ab3.png)

alexdoan102:
Set the attached [avalaraRequest_ObjectDump.txt] file for Request dump of object.

I am still getting invalid address (even with region = PA)

[avalaraRequest_ObjectDump.txt](https://github.com/VirtoCommerce/vc-community/files/292547/avalaraRequest_ObjectDump.txt)

alexdoan102:
I manually calculated the address/tax in the avalara Tax portal...and it works.... but calling through your service does not work.

![avalaramanualcalc](https://cloud.githubusercontent.com/assets/73307/15696819/356fdeae-2782-11e6-9e4b-bba608c6085e.png)

aartiomov:
AvaTax modified the Postal code "19142" -> "19142-2409". Can you provide the postal code Avalara has set after validation and try with it?
Anyway will add tests to the module in order to keep it up to date with the later releases.

alexdoan102:
new dump file of request.. ==> changing to use long zipcode  "19142-2409" , still same error..address invalid.

[...]

PR #479, opened 2016-05-31
fix [shipping address & customerId] to the taxevalcontext  (for store…

alexdoan102:
Fix for Bug #453.  This changeset updates the ShippingAddress and the CustomerId/Name for the TaxEvaluation Context.  @tatarincev asked me to fix this and create a pull request for him.

The changeset is https://github.com/alexdoan102/vc-community/commit/4c3d05f8f098de8a30d8d2b5727aa7f7d5deee48


tatarincev:
I think as temporary solution it will be enough, despite the fact that it will not work for multiple shipments with different addresses:)

PR #566, opened 2016-07-29
Updated connection strings and appSettings configuration

Esbob:
I've updated the ARM template in order to make it compatible with service principal (SPN) based authentication, which I use in Visual Studio Team Services to create VC Azure deployments. Without these changes, some weird authorization errors occur when the deployment is trying to update web.config, which tells me that the previous API version is not fully compatible with SPN.
The configuration updates are based on David Ebbo's ARM template sample found here:
https://github.com/davidebbo/AzureWebsitesSamples/blob/3ab2e9ff14c66719271a62c1ba8213c5258c7a6e/ARMTemplates/WebSiteManyFeatures.json

### DIFF
--- a/azuredeploy.json
+++ b/azuredeploy.json
@@ -150,40 +150,12 @@
                 {
                     "type": "config",
                     "name": "web",
-                    "apiVersion": "2014-04-01",
+                    "apiVersion": "2014-11-01",
                     "dependsOn": [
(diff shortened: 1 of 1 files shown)


Woland2k:
Thanks for the PR, we'll review it now. Also thanks for signing contributor agreement, you were invited to join our Contributor team.

PR #585, opened 2016-08-06
Added web.config setting VirtoCommerce:ModulesDataSources as input parameter

Esbob:
This commit makes it possible to control which version of the modules.json file (containing module versions) is being used for module installation after an Azure deployment. 
Currently, the latest version of the file is used by default, which is OK if you always deploy and install the latest version of VC. But if you need to deploy a particular version of VC platform + storefront + modules to your environments, you need to be able to "freeze" module versions as well. Since modules are installed after Azure deployment and do not have their own ARM template, I figured this commit was the easiest way to introduce proper module versioning. 
For instance, I will give the following value to the modulesDataSources input parameter if I need the module versions available on July 28th of this year:
https://raw.githubusercontent.com/VirtoCommerce/vc-modules/288b8717bd512bb826981c03bf33f533ee853206/modules.json

### DIFF
--- a/azuredeploy.json
+++ b/azuredeploy.json
@@ -54,6 +54,10 @@
                 "2"
             ],
             "defaultValue": "0"
+        },
+        "modulesDataSources": {
+            "type": "string",
+            "defaultValue": "https://raw.githubusercontent.com/VirtoCommerce/vc-modules/master/modules.json"
         }
     },
     "variables": {
@@ -204,7 +208,8 @@
                         "[resourceId('Microsoft.Web/Sites', parameters('siteName'))]"
                     ],
                     "properties": {
-                        "VirtoCommerce:AutoInstallModuleBundles": "[variables('moduleBundles')[parameters('installModuleBundle')].configValue]"
+                        "VirtoCommerce:AutoInstallModuleBundles": "[variables('moduleBundles')[parameters('installModuleBundle')].configValue]",
+                        "VirtoCommerce:ModulesDataSources": "[parameters('modulesDataSources')]"
                     }
                 },
                 {


Woland2k:
Great addition. I think you are spot on with automatic deployment of modules.

artem-dudarev:
If you freeze the modules list then you won't see updates in the module manager.
Probably we should use different setting for auto installation data source, say VirtoCommerce:AutoInstallModuleDataSource.

esbenvisfeldt:
I actually don't want the updates :-)

My use case is that I need to upgrade my existing Azure VC installation and I would like to do the following:

1. I start by deploying the latest version of VC (say v2.0) from GitHub to my dev environment. After one month, I have adapted my custom-made module to the new features available in VC v2.0 and I'm ready for test.

2. I now want to deploy VC to my test environment, but I can't take latest VC (say v2.1) from GitHub because new major changes were introduced which my module doesn't support yet. Instead, I deploy VC v2.0 and I set the modulesDataSources parameter to point to v2.0 of the modules.json file.

3. When I'm done testing, I repeat step 2 for my production environment.

My main concern here is to keep one and the same version of VC throughout my environments - this is the only way I can assure that when VC hits my production environment, it is tested and working. I will never update VC from within the VC platform manager or sync my deployment from within the Azure Web App settings panel because I don't control the VC source code repository and I can't freeze it. I could fork vc-platform, vc-storefront and vc-modules (which I actually have been forced to do because of the issues addressed in this pull request), but I'd rather avoid the hassle of doing this every time I need to upgrade my VC installation.

Esbob:
I actually don't want the updates :-)

My use case is that I need to upgrade my existing Azure VC installation and I would like to do the following:

1. I start by deploying the latest version of VC (say v2.0) from GitHub to my dev environment. After one month, I have adapted my custom-made module to the new features available in VC v2.0 and I'm ready for test.

2. I now want to deploy VC to my test environment, but I can't take latest VC (say v2.1) from GitHub because new major changes were introduced which my module doesn't support yet. Instead, I deploy VC v2.0 and I set the modulesDataSources parameter to point to v2.0 of the modules.json file.

3. When I'm done testing, I repeat step 2 for my production environment.

My main concern here is to keep one and the same version of VC throughout my environments - this is the only way I can assure that when VC hits my production environment, it is tested and working. I will never update VC from within the VC platform manager or sync my deployment from within the Azure Web App settings panel because I don't control the VC source code repository and I can't freeze it. I could fork vc-platform, vc-storefront and vc-modules (which I actually have been forced to do because of the issues addressed in this pull request), but I'd rather avoid the hassle of doing that every time I have to upgrade my VC installation

Woland2k:
It still will be good to see what updates are available, that is why different configuration property. You can still disable automatic updates in the backend.

Esbob:
If you feel that this pull request is insufficient and needs more code, I understand. The most important for me is that we find some kind of solution to the versioning issue I've described.

Woland2k:
The pull request is fine and thanks for bringing it up as I think it was overlooked. Just need to handle the case with allowing updates and we'll make those changes into the platform code.

artem-dudarev:
I think 'Modules Data Source' parameter in the deployment wizard would be very confusing for most people.
It is possible to freeze the data source manually before starting the platform:

1. Deploy the desired version of platform with Deploy to Azure button. Don't navigate to the deployed web app.
2. In the Azure web app settings add VirtoCommerce:ModulesDataSources with URL to the required version of modules.json
3. Navigate to the deployed web app

Esbob:
I agree - the name might confuse people, but that name could be changed to something else.

I'm not using the "Deploy to Azure" button - instead I've set up automatized ARM template deployment from within Visual Studio Team Services, and my goal is to have only one button to push to get a full deployment of VC platform and storefront, all included.
In that context, your step 2, which is a manual step, is the reason why I want to introduce this pull request in the first place :-)

What I've been trying to tell you through my posts is that after you've split up VC into separate pieces, it's more difficult to install and versioning is more complex (good but also bad). In my opinion you need to address this and consider how to make it easier to setup automatic deployment by improving the module versioning mechanism and minimizing manual steps, such as connection string copying (I'm thinking about CmsContentConnectionString) and API Hmac key generation. In my ideal world, after I deploy the ARM templates for VC platform and storefront, no more configuration is needed.

We should probably drop this pull request - I think I will use a PowerShell script to configure the VirtoCommerce:ModulesDataSources setting instead.

PR #645, opened 2016-10-09
added mail attachments

pierresetteskog:
just tested with sendgrid :)


tatarincev:
You forgot to add persistence for attachments, because notification sending may be delayed. And better store original attachment body in VC asset system and store within attachment as link and load it only when sending.

pierresetteskog:
Great I never scheduled the mail sending and missed that. I think its a great functionallity for i.e. sending invoice pdf's etc :)

tatarincev:
Yes, I agree is a valuable feature.

PR #669, opened 2016-11-25
Leverage Browser Caching - 7days

OlegoO:
Leverage Browser Caching - 7days 
Setting Cache-Control on Azure Blobs can help reduce bandwidth and improve the performance by preventing consumers from having to continuously download resources. 
More Info:
https://developers.google.com/speed/docs/insights/LeverageBrowserCaching
https://azure.microsoft.com/en-us/resources/samples/storage-blobs-powershell-cache-control/


tatarincev:
May be better will be do not code these CacheControl property in the code and use Azure blob  settings instead?

OlegoO:
Azure doesn't provide ability to use set CacheControl property. You should set it manually for each file.

PR #725, opened 2017-02-06
Fix for swagger ui behin https reverse proxy

pierresetteskog:
If virto is behind a reverse proxy with https the swagger ui will fail and generate http url's 

according the solutions in this threads adding following headers in the proxy should fix it.
X-Forwarded-Proto
X-Forwarded-Host
X-Forwarded-Port

https://github.com/domaindrivendev/Swashbuckle/issues/296
https://github.com/springfox/springfox/issues/1330

to quickly check my solution test with postman
image

### DIFF
--- a/VirtoCommerce.Platform.Web/App_Start/BundleConfig.cs
+++ b/VirtoCommerce.Platform.Web/App_Start/BundleConfig.cs
@@ -15,7 +15,7 @@ namespace VirtoCommerce.Platform.Web
 
         public static void RegisterBundles(BundleCollection bundles)
         {
-            //BundleTable.EnableOptimizations = true;
+            BundleTable.EnableOptimizations = false;
 
             #region CSS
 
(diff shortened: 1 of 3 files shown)


pierresetteskog:
or solve with javscript always use the same schema as window.location.protocol.
Or make it possible to change with unity.

tatarincev:
Could you explain in more detail about the core of the problem?

pierresetteskog:
our virto is behind a proxy which handles the https and sends the request further with http. the swagger ui thinks its http and generates a documentation page with http.
image

If i change to https in the input box in the picture the page will be loaded but still generate api urls for http and you cant try the api.

image

tatarincev:
Thanks for PR, we will apply it manually little later.

pierresetteskog:
I added a commit which is better and tested more in our environment. Method renamed to better describe purpose, some comments, rewrite of host port also. Please note this functionality only kicks in if the special proxy headers exists, so safe for current solutions.
more ideas whas taken from here:
https://github.com/domaindrivendev/Swashbuckle/issues/705

tatarincev:
Thanks a lot for your contribution!

PR #854, opened 2017-04-10
Platform side support for Invoice notifications and PDF export

miguelhasse:


tatarincev:
Templates preview functionality will be significant change soon. Almost all code which respond to preview will be remove. We have plans to use whole object JSON as primary data source for any template instead of use list of scalar properties.

miguelhasse:
@tatarincev what I did here was to enable notification objects with the ability to initialize themselves given the test parameters... this allows, for example, sending order notifications by specifying just the order number.

tatarincev:
Yes, I'm totally understand what you did. We will count this requirement in our work plan. Thanks for contribution.

Issue #1116, opened 2017-08-10
Platform notifications: Allow to use objects with any types in Liquid templates is not just an derived from IEntity


asvishnyakov:
Client question:
`
Hi guys, i'm trying to display the addresses on the order in the "order created" notification in a email. I get this error: Liquid syntax error: Object 'VirtoCommerce.Domain.Commerce.Model.Address' is invalid because it is neither a built-in type nor implements ILiquidizable. How do i iterate the addresses? I would like to show the shipping and invoice address in the email.
`
### Actual behavior

Get an exception when generating a notification:
` Liquid syntax error: Object 'VirtoCommerce.Domain.Commerce.Model.Address' is invalid because it is neither a built-in type nor implements ILiquidizable`

### Steps to reproduce via unit test 
Define new template notification with liquid template contains interpolation expression for object with ValueObject type.
```
{% for address in addresses %}
 {{ address.name }}
  {% endfor %}
```
Try to resolve template by call  `LiquidNotificationTemplateResolver.Resolve(notifcation)`

### Technical information:
Type  respond for rendering  notification templates
https://github.com/VirtoCommerce/vc-platform/blob/master/VirtoCommerce.Platform.Data.Notifications/LiquidNotificationTemplateResolver.cs#L45

PR #1123, opened 2017-08-11
Platform notifications: Allow to use objects with any types in Liquid templates is not just an derived from IEntity #1116

megafreeman:


asvishnyakov:
I believe you should leave only one RegisterTypeAsDrop and inherit IEntity from ILiquidizable

megafreeman:
Some edits and improvements were made in accordance with the comments @asvishnyakov @alex765022 @tatarincev

tatarincev:
`IValueObject` marker interface should be inherited only by ValueObject type. And the code that is responsible for the making of ILiquidizable must be supplemented with new `IValueObject` interface check.

megafreeman:
Moved to https://github.com/VirtoCommerce/vc-platform/pull/1127

PR #1127, opened 2017-08-14
Platform notifications:  added support to use objects inherited from IValueObjects as membr of IEntity objects or as independent object in notification templates #1116 

megafreeman:


megafreeman:
Implemented all requirements and suggestions @tatarincev @asvishnyakov

PR #1137, opened 2017-08-22
Allow to fill in azure storage account name

zhangpeng-kooboo:


artem-dudarev:
I think it will be quite inconvenient for a user to fill one more field without a default value and without hint about restricted characters.
Need to create an issue and hold a discussion.

zhangpeng-kooboo:
@artem-dudarev yes, but azure storage name have a limited length, and most of cases for a site name will contains company/project name and environment then the application name.

Issue #1114, opened 2017-08-08
Notifications do not work


asvishnyakov:
Fresh Azure deployment.
Version info: 
- Browser version: Chrome 60
- Platform version: 2.13.14
- Module version: Order 2.16.4 (for example)

### Expected behavior
Notification must be created & sent, when your action is triggered - for example, when you cancel order.

### Actual behavior
No notification created.

### Steps to reproduce
1. Make a fresh deploy of platform to Azure
2. Configure admin with sample data
3. Try to cancel order

This issue is not caused by Order module, because platform is not react to any other default triggers. Excluding new user registration notification, storefront is also affected.

tatarincev:
This commit https://github.com/VirtoCommerce/vc-module-core/commit/b6ca76d74f8a5646ed782eaa54533974dedd15fa  violated binary compatibility with the previous version, and cause next error in `OrderNotificationObserver` type
`method not found : 'System.Collections.Generic.ICollection`1<System.String> VirtoCommerce.Domain.Customer.Model.Member.get_Emails()'.`

artem-dudarev:
Fixed in https://github.com/VirtoCommerce/vc-module-order/releases/tag/v2.17.0

PR #1145, opened 2017-08-24
Fix notifications display

Skrypt:
Sent date was not displaying anymore and I removed the Order By "SentDate" since it should always be ordered by CreatedDate DESC by default.


asvishnyakov:
#1114

PR #1164, opened 2017-09-19
Username made editable

ksavosteev:


tatarincev:
Visual bug with username edit.
How to reproduce:
1. Edit username for an account
2. Click save
3.  User detail blade will lose context and become without data (all input fields are empty)

User lock-unlock UI improvements:
Instead of using  command in toolbar for unlocking user I'm would to suggest use checkbox for this operation 
image

tatarincev:
Merged
https://github.com/VirtoCommerce/vc-platform/commit/f990cff97ffc4167795ea11368935542c4727359

PR #1184, opened 2017-10-20
Dynamic property type image

ksavosteev:


VirtoCommerce.Platform.Web/Scripts/app/assets/blades/asset-upload.tpl.html
```diff
@@ -15,7 +15,8 @@
 
                     <!-- 2. ng-file-select | ng-file-select="options" -->
                     <div class="form-input __file">
-                        <input type="file" multiple uploader="uploader" nv-file-select style="z-index: 0;" tabindex="-1">
+                        <input type="file" ng-if="!blade.isImageMode" multiple uploader="uploader" nv-file-select style="z-index: 0;" tabindex="-1">
+                        <input type="file" ng-if="blade.isImageMode" uploader="uploader" nv-file-select style="z-index: 0;" tabindex="-1" accept="image/*">
```
tatarincev: Why not add Accept option for Asset uploader control instead of using special isImageMode? I guess is will be more flexible.

tatarincev:
* Image values don't saved ( Add image property to any type -> Upload image -> Click save object -> Open object again)
* Need to automatically close  the asset upload blade  when image uploaded
* Open the asset upload  blade on click by the 'Empty image' tile
* Open image in the new browser tab on click by the not empty image tile.

ksavosteev:
@tatarincev resolved those issues

PR #1185, opened 2017-10-20
Hangfire priority queus and options

WimVergouwe:
Non-intrusive Hangfire queues and additional options. Needed for  https://github.com/VirtoCommerce/vc-module-core/pull/16


VirtoCommerce.Platform.Core/Common/ICancellationToken.cs
```diff
@@ -0,0 +1,7 @@
+namespace VirtoCommerce.Platform.Core.Common
```
tatarincev: Why do we need our own ICancellationToken abstraction?  Why we can't use the default CancellationToken instead?

VirtoCommerce.Platform.Core/Jobs/JobPriority.cs
```diff
@@ -0,0 +1,9 @@
+namespace VirtoCommerce.Platform.Core.Web.Jobs
```
tatarincev: I wonder to know, how to then configure Hangfire priority of resources for this named queues? Any link to documentation will be appreciated!

VirtoCommerce.Platform.Core/Common/CancellationTokenWrapper.cs
```diff
@@ -0,0 +1,19 @@
+using System.Threading;
```
tatarincev: Could you explain why we need it?

VirtoCommerce.Platform.Core/Jobs/JobPriority.cs
WimVergouwe: Hangfire will process jobs in order of the queues: http://docs.hangfire.io/en/latest/background-processing/configuring-queues.html

VirtoCommerce.Platform.Core/Common/CancellationTokenWrapper.cs
WimVergouwe: I've added comments inline in code.

VirtoCommerce.Platform.Core/Common/ICancellationToken.cs
WimVergouwe: I've added comments inline in code.

tatarincev: Thanks. I've looked at this article, but I missed this line  'The order is important, workers will fetch jobs from the critical queue first, and then from the default queue.'

VirtoCommerce.Platform.Core/Jobs/JobPriority.cs
tatarincev: Thanks. I've looked at this article, but I missed this line  'The order is important, workers will fetch jobs from the critical queue first, and then from the default queue.'

VirtoCommerce.Platform.Core/Common/CancellationTokenWrapper.cs
tatarincev: So then if it required only for Hangfire jobs, maybe better will be moving this wrapper and abstraction to Jobs namespace?

WimVergouwe: I'm sorry, I don't see how that could work, we need it in the indexing manager right?

tatarincev: Well, I saw something similar in your PR, it may be just duplicate. If yes just need to remove it.

WimVergouwe: This wrapper is to support System.Threading.CancellationTokens, the other wrapper is for Hangfire and is placed in the job namespace.

tatarincev:
Looks all ready to merge, I'll try to release today, thank you!

Issue #25, opened 2015-01-26
ESWorker Role doesn't work


kodermax:
Установил на azure движок, но почему то роль elasticsearch не работае http://motul-search.cloudapp.net:9200 не отвечает на этот порт, как быть?

Woland2k:
Are there any errors in azure role? Can you check the status of it?

kodermax:
Only these 
![err4](https://cloud.githubusercontent.com/assets/590985/5905533/cb991530-a5a1-11e4-93ff-3c6694f70aab.png)

Woland2k:
You can change the log settings for the service to get more information. But speaking of service, what is the status of it, does it say "Running" on azure website itself?

kodermax:
Yes it's running. How can y reinstall this role on azure?

kodermax:
New errors.
![err5](https://cloud.githubusercontent.com/assets/590985/5905933/e2c6193a-a5a4-11e4-9936-2b440e6489d5.png)

Woland2k:
Try expanding application logs. Simply click on them and select view all. Which version of VC are you running? 1.12?

kodermax:
also erros

RoleInstanceId: ESWorkerRole_IN_0; DeploymentId: 37cef048b7ac4475ae901198d99fef67; RoleName: ESWorkerRole
Checking VM Type
File HandlerEnvironment.json not found. Extension set to PAAS
Store Size: 8192
Reading XML configuration file
Will use thumbprint 78AD2563BA50E49635B796D5454C965F143F1626
Decrypting private configuration
Thumbprint: 78AD2563BA50E49635B796D5454C965F143F1626
Stopping the running plugin, if any
No DiagnosticsPlugin is running.
Killing any DiagnosticsPlugin that might remain.
Killing any MonAgentHost that might remain.
Diagnostics plugin stopped successfully
Resetting DiagnosticPlugin
Starting DiagnosticPlugin.exe DiagnosticPlugin-ShutdownEvent DiagnosticPlugin-TotalShutdownEvent -wadVer1v3
Starting DiagnosticsPlugin
Checking VM Type
File HandlerEnvironment.json not found. Extension set to PAAS
RcfFileName: C:\Config\37cef048b7ac4475ae901198d99fef67.37cef048b7ac4475ae901198d99fef67.ESWorkerRole_IN_0.2.xml
[...]

PR #1212, opened 2017-11-08
add dependenc infinite scroll for grid #25

n2pro:


n2pro:
Use of infinite scrolling without modification of the platform

Issue #1117, opened 2017-08-10
Setup database connection timeout directly in code, based on connection string


asvishnyakov:
Version info: 
- Platform version: All, 2.13.14 affected too

### Expected behavior
Platform db connection must fail after 420 seconds, as we setup in default connection string.

### Actual behavior
Platform fail after ~30 min. This is default connection timeout in EF.
It's caused by EF, because it has bug, which prevent EF to use timeout from connection string. So we must read timeout from connection string and setup it explicitly in code.
https://stackoverflow.com/a/6234593

### Steps to reproduce
Edit your connection string and make it wrong (for example, specify non-existing SQL instance). Then just wait ~30 min.

eMazeika:
### Observations on Actual behavior
The Connect Timeout actually is taken from connection string. Just it behaves in a strange way. I was changing the value in connection string and waiting for initial page to timeout.
Connect Timeout -> Page times out in:
3     -> 	13s
4     -> 	27s
5     -> 	29s
6  	  ->	32s
10    ->	46s
15    ->  	1.8 min
120   ->	16.5 min
420	  ->	1.1 hour

Each value was tested several times, receiving very similar timeout times (using my PC). Tried setting the timeout value explicitly in code, but w/o any effect.
BTW, there is no such thing as "default connection timeout in EF"; the default connection timeout in SqlClient is [15 seconds](https://msdn.microsoft.com/en-us/library/system.data.sqlclient.sqlconnection.connectiontimeout.aspx).

tatarincev:
Released https://github.com/VirtoCommerce/vc-platform/releases/tag/v2.13.37

PR #1213, opened 2017-11-10
Features/migration timeout

Elyseum:
Adding option to make the DBMigration SQL CommandTimeout configurable through appsettings because the default of 30 seconds was insufficient for large sets of data.


Elyseum:
Sure, agreement accepted.

It's not clear to me what you mean with old EF bug. Use a connection string property to set the command timeout? We chose a separate setting because it's targeted for migration explicitly: these jobs potentially run a lot longer than default user requests.

tatarincev:
@Elyseum I mean this EF bug https://github.com/virtocommerce/vc-platform/issues/1117
But now I understand you want to have special SQL timeout settings only for migration queries.

PR #1221, opened 2017-11-20
Override templates for metaproperty types

ksavosteev:


ksavosteev:
@tatarincev removed valueInputs.tpl.html, moved templated back to the old file, removed directive templateUrl

PR #1225, opened 2017-11-22
Change URI when open blob for read using CDN

megafreeman:
We get an error in ImageTools module when trying to generate thumbnails for uploaded images. Azure CloudBlobClient does not work correctly when accessing the image via CDN Url. So, in this fix, we replace CDN Url with base storage Url when open blob for reading.


megafreeman:
@tatarincev suggested to use storage base URI for all cases when open blob for reading

PR #1229, opened 2017-11-27
Features/support neutral culture

ArvidBauters:
Using a neutral language instead of a specific language in Virto causes multilingual fields to not show flags. Flags are determined by taking the country part of a specific Language. This change replaces the flag by the two letter culturecode in case of neutral languages. For specific languages, nothing will change.

Also see PullRequests on vc-module-catalog and vc-module-core for related changes.


ArvidBauters:
Hi,
Related PR on vc-module-catalog has been approved.
See: https://github.com/VirtoCommerce/vc-module-catalog/pull/79

 Could anyone have a look at this one as well, as this PR contains the necessary css to make it look good.

PR #1245, opened 2018-01-10
add copy and move for azure provider #2

n2pro:


tatarincev:
Absolutely incorrect implementation.  
You can't  add new methods to one of IBlobStorageProvider interface implementation without adding these methods to the abstraction interface, without it clients who works with this abstraction will nothing to know about this new methods.
Well, it does not matter,  it's not right to do it on the platform, move operations can be implemented by using already exist in IBlobStorageProvider methods  (copy and delete) and shoudl be done in the vc-content module in this concrete type 
https://github.com/VirtoCommerce/vc-module-content/blob/master/VirtoCommerce.ContentModule.Data/Services/AzureContentBlobStorageProvider.cs#L26

Issue #83, opened 2015-10-07
Move githubrepository to token instead of username password


Woland2k:
Look here for info: https://github.com/octokit/octokit.net/blob/master/docs/oauth-flow.md

PR #1250, opened 2018-01-16
removed excess leading slash #83

andshk:


tatarincev:
Done by different way https://github.com/VirtoCommerce/vc-module-catalog/pull/102/files

PR #1290, opened 2018-02-27
fix height grid

n2pro:
ytdxvdsjs2
If the directive does not pass the attribute, it is still used,
If you pass the attribute, then the height of the grid is calculated depending on the space occupied in the blade.
In the directive, we pass the selector in which the grid is located
```
<div class="blade-content __medium-wide">
    <div class="blade-inner">
        <div class="inner-block">
            <div class="part-1"></div>
            <div class="part-2">Here UI-GRID <div ui-grid="gridOptions" ... ui-grid-height=".part-2"</div>
        </div>
     </div>
</div>
```


tatarincev:
Released https://github.com/VirtoCommerce/vc-platform/releases/tag/v2.13.33

PR #1297, opened 2018-03-05
added stream upload file to UploadAsset

n2pro:


n2pro:
You can check the download of files using swagger UI
uploadfile

PR #1310, opened 2018-03-19
Allow unregistering of notification types.

vanzegbroecke:
Allows unregistering of notifications: for example when you don't want to send several mails while creating an order.

Related to https://github.com/VirtoCommerce/vc-module-order/pull/33

As discussed on Gitter: lots of changes are coming to the notification module, so this is a fix until that has been done. It will be better if 'IsActive' is actually used or there is another property to use for it, as that could then be used in the search notifications criteria. Right now, if you used to have a notification of a certain type, but it has been unregistered, it will show as an empty row in the notification history.. That could be fixed by using IsActive.


vanzegbroecke:
@tatarincev The order notification observer sets everything to active after getting the notifications, though, so that would still have to change. It also seems kind of weird to have to create a type, just to be able to remove it?

tatarincev:
Yes, I've seen the order module, and IsActive = True is forced for all notifications, then there is no chance :)

PR #1318, opened 2018-03-29
Added api for lockout user

n2pro:


tatarincev:
Redundant changes, these functionality can be achieved without these changes

PR #1323, opened 2018-04-16
changed swagger config to  work behind azure waf a

pierresetteskog:
I.e. this makes it possible to make the waf handle https and forward the traffic to the webapp with http (and even other domain).

This is an fix on an old fix but now its not case sensitive and works with azure waf.

Without this fix the swagger ui page will be generated with http urls even when called with https because of in our setup the waf is handling https.

T


pierresetteskog:
please let me know if any questions! :)

Issue #1349, opened 2018-06-15
Implement DB storage for platform Push notifications


tatarincev:
Problem:
Now we have only InMemory build-in storage for all the platform push notifications and it leads to problems when platform running on multiple instances.
Solution:
Need to replace current 'InMemory'  PushNotificationManager with new implementation which would use DB for store pushNotifications records.
To do that, need to extend `IPlatformRepository` with new table `PlatformPushNotifcation` and implement the follow classes  structure
![image](https://user-images.githubusercontent.com/7566324/41465109-f902675a-709c-11e8-811c-71749e141e9c.png)

All previous WEB API methods should be preserved for backward compatibility with already exists UI

Woland2k:
Shouldn't we use some queueing service instead? There is one available in azure.

tatarincev:
The use of additional services such as `Azure.ServiceBus` at the moment significantly complicates the platform primary configuration and deployment. Therefore, using a DB storage for these purposes looks as the simplest solution to this problem.

Woland2k:
it won't complicated things much in azure (it is a simple queue service, not servicebus), and it is a better approach for notifications than database.

tatarincev:
There is PR https://github.com/VirtoCommerce/vc-platform/pull/1350 about this feature need to verify it and  test

PR #1350, opened 2018-06-19
Push notification db storage

n2pro:


avas:
This PR is related to the issue #1349. The solution provided by @n2pro mostly worked, I've just added few modifications:
* In the VC platform, push notifications are mostly implemented as custom classes which derive from `PushNotification` class and extend it with some additional properties. Previous implementation was only saving `PushNotification` properties, losing the type of the notification and all its custom properties. To fix it, I added source notification serialized to JSON and its assembly-qualified type to the `PushNotificationEntity`.
* To achieve that, I also needed to change push notification instantiation code in the `PushNotificationService` class. Previous implementation used the same type for all notification; current implementation attempts to instantiate actual source class (using the value of `PushNotificationEntity.AssemblyQualifiedType` property). If the actual source push notification class is not available anymore, current implementation uses the `PushNotification` class as the fallback option.
Please see the `PushNotificationService.ConvertEntitiesToModels` for further details.
* Fixed push notifications sorting on the notifications history blade not working;
* Fixed push notification about installed/updated/deleted modules not showing up on the notification history blade.
* Minor non-breaking renames and codestyle-related tweaks.

Current implementation properly handles module updates (tested by updating CSV catalog import module from 1.0.1 to 1.1.0). It can also display (sort of) push notification from some module after uninstallation of that module:
image

avas:
Oh, and one more concern: with the current implementation the platform does not provide any way to clean up the `PlatformPushNotification` table, so it will grow infinitely. We should probably add some way to remove old entries (like some automated job removing old entries and/or some button on the push notification history blade).

yecli:
Implemented in v3 another way

PR #1366, opened 2018-07-09
Bug fix subscriptions count

Tilesto:
screenshot_1
screenshot_2
screenshot_3


VirtoCommerce.Platform.Web/Content/themes/main/sass/blade-constructor.scss
```diff
@@ -0,0 +1,44 @@
+﻿/*
```
n2pro: See if other changes are needed for other modules

PR #1405, opened 2018-08-06
Design refactoring

Tilesto:
Project - folder "Design"
Refactoring of css colors.
All colors using changes to css-variables.
Beta theme-color-change realised.

Preview theme change: http://recordit.co/6zxPqX63iq


asvishnyakov:
@tatarincev  Design folder is outdated and not compatible with current platform UI. You need to actualize styles & layout first

Woland2k:
Also it should be possible to set blade menu and blade top backgrounds to be different. I'd like to have blade menu to have a white background instead of solid color.

PR #1408, opened 2018-08-09
Dot not mark tags-input with error until you interact with control

asvishnyakov:
### Problem
Currently, if you will set min-tags="1", for example, then tags-input will have red border even if you first time open blade and did nothing (i.e. form is untouched)

### Solution
Dot not mark tags-input with error until you interact with control

### Proposed of changes
If tags-input have ng-untouched class, remove red border (box-shadow used for that in tags-input css)


yecli:
Released in [v2.13.42](https://github.com/VirtoCommerce/vc-platform/releases/tag/v2.13.42)

Issue #845, opened 2017-04-05
Different behavior for IBlobStorageProvider implementations


asvishnyakov:
`AzureBlobProvider` and `FileSystemBlobProvider` has different behavior for `IBlobStorageProvider.OpenWrite` method implementation.
* `AzureBlobProvider` return *write-only* stream from Azure blob service
* `FileSystemBlobProvider `create new file with *read/write* permissions

Because of that, code what works on developer local machine may fail in Azure. It's simple to fix this if there is no side-effects on platform and modules: just replace `File.Open(filePath, FileMode.Create)` in [FileSystemBlobProvider.cs](https://github.com/VirtoCommerce/vc-platform/blob/f75b0d2625fe3f5ad6d3866687887072d72c2b8e/VirtoCommerce.Platform.Data/Assets/FileSystemBlobProvider.cs#L86) with `File.Open(filePath, FileMode.Create, FileAccess.Write)`

OlegoO:
@asvishnyakov I approve fix but you should create unit tests

PR #1417, opened 2018-08-20
#845 Change FileSystemBlobProvider.OpenWrite permissions

VitalyTartynov:
### Problem
Different behavior for IBlobStorageProvider implementations described in #845 issue.

### Solution
Solution from discussion in #845.
Now FileSystemBlobProvider returns write-only stream. Added unit-test for checking this logic.

### Proposed of changes
Add `FileAccess.Write` parameter in creation file stream.


VirtoCommerce.Platform.Tests/Assets/BlobStorageProviderTests.cs
```diff
@@ -0,0 +1,43 @@
+using System;
+using System.IO;
+using VirtoCommerce.Platform.Data.Assets;
+using Xunit;
+
+namespace VirtoCommerce.Platform.Tests.Assets
+{
+    public class BlobStorageProviderTests
```
tatarincev: Please add `[Trait("Category", "Unit")]` attribute to test class, this attribute tells our CI to run it for each build.

PR #1418, opened 2018-08-20
ToNullable should return null instead of default(T)

artem-dudarev:
### Problem
When trying to save `null` value for a dynamic property value of type `DateTime` the `null` value is converted to `0001-01-01T00:00:00.0000000`, which cannot be saved to the database.

### Solution
Don't convert `null` to `default(T)`


asvishnyakov:
@tatarincev High priority for our team too.

tatarincev:
Released https://github.com/VirtoCommerce/vc-platform/releases/v2.13.36

Issue #1433, opened 2018-08-31
Allow users to change their password when they sign in for the first time


tatarincev:
**Functional requirements:**

As Administrator I want's to (reset/set temporary) password for selected user which it must  change on they first sing in 
As Administrator I want's  see  the password strength for a new password
As Administrator I want's see a validation error if entered password doesn't match the system password strength policy
As User I must change my temporary password on first sign in in the system with use the curent system password strength policy

**Technical details:**
- Need to change current "Use password change" blade 
![image](https://user-images.githubusercontent.com/7566324/44901338-670bc100-ad07-11e8-99f6-d4646168ece9.png)
to the new template
![image](https://user-images.githubusercontent.com/7566324/44901290-42afe480-ad07-11e8-968b-0a799ec99820.png)

- For a password, strong checking need to provide some new API service which will be based on the current **ASP.NET Identity** password strong settings to validate and calculate strength for given password.
Implementation can be taken from this article:
https://www.ryadel.com/en/passwordcheck-c-sharp-password-class-calculate-password-strength-policy-aspnet/

- For validation errors display need to use the folowing statements localized from resources 
Passwords must have at least one lowercase ('a'-'z')
Passwords must have at least one uppercase ('A'-'Z')
Passwords must have at least one digit ('0'-'9')
Passwords must have at least one non alphanumeric character

- Need to use `ApplicationUserExtended.PasswordExpired` property as a primary flag to determine what user must change its password after sign in  

- The platform already has a logic to change user password on first sign in need just checked it out what it works for this case
https://github.com/VirtoCommerce/vc-platform/blob/master/VirtoCommerce.Platform.Web/Scripts/app/security/blades/account-resetPassword.js

![image](https://user-images.githubusercontent.com/7566324/44902768-20b86100-ad0b-11e8-9a3e-dc04e561457d.png)

avas:
The blade currently looks like this:
![- -2018-09-07-12 16 08](https://user-images.githubusercontent.com/1835759/45201148-46f77700-b29e-11e8-8485-7bf58b428a4a.gif)

The "Make this user change their password..." switch does not work yet - I'll make it working during the current issue part.

tatarincev:
Released https://github.com/VirtoCommerce/vc-platform/releases/v2.13.37

PR #1434, opened 2018-09-07
1433: changed look of password reset blade, added password validation hints

avas:
Changes made for issue #1433:
* Added `IPasswordCheckService` interface implemented by `PasswordCheckService` class that is tested by `PasswordCheckServiceTests`. It validates password strength and returns `PasswordValidationResult` that helps to decide:
    * Is the password valid? Can it be used as account password?
    * If not, what rules it violates?
* `SecurityController`: added `ValidatePasswordAsync` method that invokes password validation (using `IPasswordCheckService`). It always returns `Ok(passwordValidationResult)`, regardless of password validation status.
* Changed look of the password reset blade: 
    * Removed input for password re-typing;
    * Added switch for resetting user password when first logging in (it doesn't work yet, but required for future parts of this issue).
    * Added password strength validation. Now the blade calls `SecurityController.ValidatePasswordAsync()` on each password change and, if the password is not valid, displays error messages.

Note that requests from blade to `SecurityController.ValidatePasswordAsync()` are not throttled. This method does not do any heavy-processing like database queries or other service requests (at least for now), so it *probably* should be OK.


VirtoCommerce.Platform.Data.Security/SecurityService.cs
```diff
@@ -266,7 +266,7 @@ public virtual async Task<SecurityResult> ChangePasswordAsync(string name, strin
             }
         }
 
-        public virtual async Task<SecurityResult> ResetPasswordAsync(string name, string newPassword)
+        public virtual async Task<SecurityResult> ResetPasswordAsync(string name, string newPassword, bool forcePasswordChange)
```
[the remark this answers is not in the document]
avas: Well, I added it for future work on issue #1433. But I agree, it is not used yet and isn't really related to the purpose of this PR. Removed it.

VirtoCommerce.Platform.Web/Security/AuthenticationOptionsBasedPasswordCheckOptions.cs
```diff
@@ -0,0 +1,39 @@
+using VirtoCommerce.Platform.Core.Security;
+using VirtoCommerce.Platform.Data.Security;
+
+namespace VirtoCommerce.Platform.Web.Security
+{
+    /// <summary>
+    /// Implementation of <see cref="IPasswordCheckService"/> that gets password validation options from
+    /// <see cref="AuthenticationOptions"/>.
+    /// </summary>
+    public class AuthenticationOptionsBasedPasswordCheckOptions : IPasswordCheckOptions
```
tatarincev: I guess better will be using already exists `AuthenticationOptions` type instead of this decorator

VirtoCommerce.Platform.Web/Scripts/app/security/blades/account-resetPassword.js
```diff
@@ -1,27 +1,71 @@
...
-            blade.error = 'Error: new passwords doesn\'t match!';
-            return;
-        }
 
+    $scope.validatePasswordAsync = function (value) {
+        var promise = accounts.validatePassword(JSON.stringify(value)).$promise.then(
+            function(response) {
+                blade.currentEntity.errors = [];
+
+                if (response.passwordIsValid) {
```
tatarincev: This long switch/case can be simplified with a one-line loop where we can use JS `Object.keys()` to get properties names and use them to choose appropriate translation key from resources

avas:
Pushed some changes to ask users to change their password after sign in:
- `AccountEntity`: added persistable `PasswordExpired` property to indicate whether user should be prompted to change their password on next sign-in. Value of this property is passed to `ApplicationUserExtended.PasswordExpired`.
- `SecurityService`
    - `ResetPasswordAsync`: added `forcePasswordChange` parameter that gets assigned to `AccountEntity.PasswordExpired`.
    - `ChangePasswordAsync`, `ResetPasswordAsync`: added code to reset `PasswordExpired` flag for user (since the previous expired password *did* change, and the new password is not expired anymore).
- `SecurityController`: added `ResetCurrentUserPassword` method that resets password of current platform user. It works same as `ResetPassword` method, but does not require `PredefinedPermissions.SecurityUpdate` permission and can be used by any authorized user.
- Changed `changePasswordDialog`: now it changes password for current user only (previous implementation have been able to change password for any user). This was necessary because previous implementation required admin privileges and thus cannot be used by regular users to change their password. The change was suggested by @tatarincev.
- Extracted password validation code from `account-resetPassword` blade to the new `platformWebApp.passwordValidationService` to reduce code duplication.
- Added password strength validation to `changePasswordDialog`.
- `app.js`: added code that redirects user to `passwordChangeDialog` after sign in if they have to change their password.

VirtoCommerce.Platform.Web/Scripts/app/security/security.js
```diff
@@ -115,18 +115,33 @@
...
                             $stateParams.onClose();
                         }
                     });
 
+                    $scope.validatePasswordAsync = function(value) {
+                        var promise = passwordValidationService.validatePasswordAsync(value).then(
+                            function (result) {
+                                $scope.passwordValidationResult = result;
+
+                                return result.passwordIsValid ? $q.resolve() : $q.reject();
```
tatarincev: Why do you use `$q.resolve() : $q.reject();` this? I think all can work without these calls. Just returns result.passwordIsValid  should be enought.

VirtoCommerce.Platform.Web/Scripts/app/security/blades/account-resetPassword.js
```diff
@@ -1,27 +1,37 @@
...
-            blade.error = 'Error: new passwords doesn\'t match!';
-            return;
-        }
 
+    $scope.validatePasswordAsync = function (value) {
+        var promise = passwordValidationService.validatePasswordAsync(value).then(
+            function(result) {
+                angular.extend(blade.currentEntity, result);
+
+                return result.passwordIsValid ? $q.resolve() : $q.reject();
```
tatarincev: Why do you use $q.resolve() : $q.reject(); this? I think all can work without these calls. Just returns result.passwordIsValid should be enought.

VirtoCommerce.Platform.Web/Scripts/app/security/security.js
avas: I did it because of `ui-validate-async` nature - it [expects](https://github.com/angular-ui/ui-validate/blob/master/src/validate.js#L108) to receive one of the following objects:
* Promise - this way control will be valid if the promise was resolved, and invalid if the promise was rejected;
* Anything else - this way `uiValidateAsync` directive will wrap it to promise and resolve/reject it if the value can or can't be implicitly converted to `true`.

If we'll replace `return result.passwordIsValid ? $q.resolve() : $q.reject()` with just `return result.passwordIsValid`, `validatePasswordAsync()` method will always return resolved promise with `passwordIsValid` as its result, because the HTTP call to password validation method always returns `Ok()`, even if password is invalid. So the password input will always stay valid, because the `ui-validate-async` directive looks only for resolution/rejection of the promise and ignores resolution result.

avas:
Added password validation to account creation blade:
image

VirtoCommerce.Platform.Web/Scripts/app/security/security.js
tatarincev: If this  function returns `result.passwordIsvalid` boolean value which  will normally be handled by `ui-validate-async`  directive in else statement  https://github.com/angular-ui/ui-validate/blob/master/src/validate.js#L111

avas: I tried that, and the password input always stays valid. Apparently, it happens because `validatePasswordAsync()` still returns promise, so it gets handled by `if` statement on [line 108](https://github.com/angular-ui/ui-validate/blob/master/src/validate.js#L108).

avas:
* `passwordValidationService`: added throttling for password validation requests. Default throttling interval is set to 100 ms, but this value can be changed if needed by changing `passwordValidationService.throttleTimeoutMilliseconds` value.
* Added `va-password-input` directive to make password validation easier. It can be used like this:

```html
<label class="form-label">{{ 'platform.blades.change-password.password' | translate }}</label>
<va-password-input password-placeholder="{{ 'platform.blades.account-resetPassword.placeholders.new-password' | translate }}"
                   password-too-weak-message="{{ 'platform.blades.account-resetPassword.validations.password-too-weak' | translate }}"
                   run-password-validation="validatePasswordAsync(value)"
                   new-password="password" />
```

It accepts following parameters:
* `password-placeholder`: placeholder text for password input;
* `password-too-weak-message`: message that will be displayed if password is considered too weak. It will be followed by error messages;
* `run-password-validation`: method binding that will be used for actual password validation. It must accept one parameter named `value` that will contain the entered password. It must return a promise that resolves to a result with following properties:
    * `passwordIsValid`: boolean value indicating if password is considered valid;
    * `errors`: array of resource keys that describe errors found when validating password;
    * If error resource strings contain any substitutable variables (e.g. `Passwords must be {{minPasswordLength}} or more characters long`), result must also contain those variables.
* `new-password`: binding to a property that will be used to store entered password value. Property value will be updated only if entered password is considered valid.

Issue #1388, opened 2018-07-26
VirtoCommerce should provide ability to use Azure Active Directory as an authentication provider


OlegoO:
**Problem:**
We use the corporative Azure Active Directory as a corporative standard. We want to use Azure Active Directory as an authentication provider for single sign in.

**User stories:**

As a manager, I want to sign in to Virto Commerce portal with Azure Active Directory account.
GIVEN An user has Azure Active Directory account
WHEN The user opens Virto Commerce portal
AND The user is not authenticated 
THEN The user should be redirected to Azure Active Directory

GIVEN The user has Azure Active Directory account
AND The user was redirected from  Virto Commerce portal
WHEN The user passes the sign in Azure Active Directory
THEN The user should be redirected to Virto Commerce portal 
AND I see Virto Commerce portal based on user role.

As an administrator, I want to manage roles and permissions in Virto Commerce portal.

As an administrator, I want to manage account in Azure Active Directory .

**Solution:**
https://docs.microsoft.com/en-us/azure/app-service/app-service-authentication-overview
https://docs.microsoft.com/en-us/azure/app-service/app-service-mobile-how-to-configure-active-directory-authentication

avas:
Steps to test this feature:
1. Register platform application in Azure Active Directory:
    1. Go to the [Azure portal](https://portal.azure.com/).
    2. Register an application: **Azure Active Directory > App registrations > New application registration**.
    3. Enter app registration details: ![image.png](https://images.zenhubusercontent.com/5b7037d390d0044746ba55b5/9fab1f3f-bf50-47b8-8167-42382430d2cb)
        * Name - just any human-readable name;
        * Type - `Web app / API`;
        * Sign-on URL - the URL of the application. It should not necessarily be URL of an Azure web site - in my case it just points to a platform on my local IIS.
    4. After creation of an application, note the `Application ID` - you'll need it on the next step.
2. Change Azure AD authentication options in `Web.config`:
    * Change `VirtoCommerce:Authentication:AzureAD.Enabled` to `true`;
    * Paste your Azure AD application ID to the value of `VirtoCommerce:Authentication:AzureAD.ApplicationId`;
    * Insert your Azure AD domain ID to `VirtoCommerce:Authentication:AzureAD.TenantId`. It can be found on the Azure portal: **Azure Active Directory > (your domain) > Properties > Directory ID**.
3. Try the feature:
    * Launch the VirtoCommerce platform and navigate to the login page.
[...]

PR #1438, opened 2018-09-21
1388: added Azure AD authentication

avas:
The authentication is made using OpenID Connect. 

**NOTE**: Some additional set-up is required to test this feature - please see the [walkthrough](https://github.com/VirtoCommerce/vc-platform/issues/1388#issuecomment-423525628) in the issue comment for further details.

Changes made for #1388 to add the external authentication with Azure AD account:
* Added some Azure AD-related properties to the Web.config and corresponding properties in `AuthenticationOptions` class. Also added code to map it together to the `Startup` class.
* Added `Microsoft.Owin.Security.OpenIdConnect` to `VirtoCommerce.Platform.Web`. It is initialized in `OwinConfig.cs`.
* Added the `ExternalLoginController` that redirects the user to the Azure AD sign-in page and creates/updates the VC platform account if needed. It uses the `ISecurityService` to work with platform accounts. All dependencies are injected with DI - I enabled it for MVC in `VirtoCommercePlatformWebBootstrapper.cs`.
* Login dialog: added link for Azure AD sign-in. It points to the `ExternalLoginController.SignIn` action.


VirtoCommerce.Platform.Web/VirtoCommercePlatformWebBootstrapper.cs
```diff
@@ -49,7 +51,7 @@ public override void Run(bool runWithDefaultConfiguration)
             base.Run(runWithDefaultConfiguration);
 
             //registering Unity for MVC
-            //DependencyResolver.SetResolver(new UnityDependencyResolver(container));
+            DependencyResolver.SetResolver(new MvcUnityDependencyResolver(Container));
```
tatarincev: Is this change necessary?

VirtoCommerce.Platform.Web/VirtoCommerce.Platform.Web.csproj
```diff
@@ -1189,6 +1204,7 @@
     <Folder Include="App_Data\Modules\" />
     <Folder Include="Filters\" />
     <Folder Include="Modules\" />
+    <Folder Include="Views\ExternalLogin\" />
```
tatarincev: Empty folder

VirtoCommerce.Platform.Web/Scripts/app/security/login/login.tpl.html
```diff
@@ -38,6 +38,11 @@
                     {{authError}}
                 </div>
             </div>
+            <div class="form-group">
```
tatarincev: I think we shouldn't display this action if  `VirtoCommerce:Authentication:AzureAD.Enabled` setting is off

VirtoCommerce.Platform.Web/Controllers/ExternalLoginController.cs
```diff
@@ -0,0 +1,200 @@
...
+            await signInManager.SignInAsync(aspnetUser, true, true);
+            await _eventPublisher.Publish(new UserLoginEvent(platformUser));
+        }
+
+        private async Task AddExternalLoginToExistingUser(ApplicationUserExtended platformUser,
+            ExternalLoginInfo externalLoginInfo)
+        {
+            var externalLogins = platformUser.Logins?.ToList() ?? new List<ApplicationUserLogin>();
+
+            var externalLogin = ConvertExternalLoginInfoToExternalLogin(externalLoginInfo);
```
tatarincev: Are you sure what this logic won't create duplicate records after each external login?

VirtoCommerce.Platform.Web/Controllers/ExternalLoginController.cs
```diff
@@ -0,0 +1,200 @@
...
+
+            var identity = externalLoginInfo.ExternalIdentity;
+            var userName = identity.FindFirstValue(ClaimTypes.Upn);
+            if (string.IsNullOrWhiteSpace(userName))
+            {
+                throw new InvalidOperationException("Received external login info does not have an UPN claim.");
+            }
+
+            var signInManager = _signInManagerFactory();
+            var externalLoginResult = await signInManager.ExternalSignInAsync(externalLoginInfo, false);
```
tatarincev: There is some code duplications in `SignInExternalUser` and `CreateOrUpdatePlatformUser` methods
```
 await signInManager.SignInAsync(aspnetUser, true, true);
 await _eventPublisher.Publish(new UserLoginEvent(platformUser));
```

 I think it can be eliminated after some  little code refactoring to the next logic flow
1.  `ExternalSignInAsync` 
2. If exist user try to add external login and update user 
3. else need to create new user with external logins
4. user = userManager.FindByLoginInfo(loginInfo)
5. signInManager.SignIn(user)
6. publish event

VirtoCommerce.Platform.Web/VirtoCommercePlatformWebBootstrapper.cs
avas: Yes, it enables DI for regular (non-WebAPI) MVC controllers. It is used to inject dependencies to the `ExternalLoginController`.

VirtoCommerce.Platform.Web/Controllers/ExternalLoginController.cs
avas: Well, it shouldn't - if the user already had an external login entry, they will sign in successfully (`externalLoginResult == SignInStatus.Success` on line 83), and this method won't get called. And even if it *will* create duplicate record and attempt to save it, it will lead to duplicate record in `dbo.AspNetUserLogins` table - such record will violate the primary key constraint, and RDBMS will not allow to save it.
image

avas:
* Refactored `ExternalLoginController` a bit to reduce code duplication.
* Reworked the code to display external login link(s) on login dialog (see 87c7148):
    * Added `SecurityController.GetExternalLoginProviders` method (available at `/externalloginproviders`), so that frontend could request available external providers;
    * Login dialog now requests external login providers and displays `Sign in with (provider name)` link for each provider. This way Azure AD sign-in link won't be displayed on login dialog when Azure AD authentication is disabled. It will also make it easier to add more external sign-in providers if needed.

VirtoCommerce.Platform.Web/Controllers/ExternalLoginController.cs
avas: Reworked the code that adds external logins. Now there won't be any external login duplications.

avas:
@tatarincev Of course, I'll do it.

avas:
@tatarincev Well, the code looks much cleaner and mostly works. But it fails in one particular case: if the VC platform account with Azure AD login is already created (e.g. manually by admin), and that user attempts to log in externally. In this case platform will attempt to create a new user account and will fail, because that account already exists - it is just missing the `ExternalLogin`.
image

Will this case be used by end users of this feature? If so, contents of the `if (externalLoginResult == SignInStatus.Failure) { ... }` block will need one more `if`.

avas:
I've made some changes to fix this error. The code works fine for me, but looks a bit uglier.
@tatarincev Can you please check changes made in 0115847?

avas:
@tatarincev Turns out that `User.Identity.Name` can be empty after external sign-in, even if that sign-in was successful. So I guess we'll really have to use `userName` extracted from user claims.
image

I also wasn't sure whether the `ApplicationUserExtended.Logins` can be null, but turns out it can't - at least I didn't achieve that. So the code is okay.

I've pushed my changes in 9912229. Can you please take a look?

VirtoCommerce.Platform.Web/Controllers/Api/SecurityController.cs
```diff
@@ -588,6 +592,24 @@ public async Task<IHttpActionResult> UnlockUserAsync(string id)
             return Content(result.Succeeded ? HttpStatusCode.OK : HttpStatusCode.BadRequest, result);
         }
 
+        [HttpGet]
```
tatarincev: I think this method  can be moved into ExternalLoginController

PR #1455, opened 2018-10-17
Delaware pro features/command timeout

WimVergouwe:
### Problem
SQL timeouts on various operations when using a large database.

### Solution
Configure the command timeout on the unit of work and pass this to the EF context.

### Proposed of changes
See PR code.


WimVergouwe:
@tatarincev , that commit sets the command timeout for the entire repository, but we need different timeouts depending on the query. This typically happens with larger databases. Do you agree?

tatarincev:
But you might do it in the place `((System.Data.Entity.DbContext)repository).Database.CommandTimeout  = value` 
Why do you need extending abstraction fro this?

WimVergouwe:
Ok that's possible too, but it breaks your abstraction, because now you depend on EF.
I'll refactor it.

WimVergouwe:
I just realized that the repository is created for each request. So I'll implement your suggestion.

PR #1457, opened 2018-10-25
CartModule, 28: Swagger JSON validation issues

avas:
This PR is a part of the VirtoCommerce/vc-module-cart#28. It helps to solve Swagger validation errors like `$ref values must be RFC3986-compliant percent-encoded URIs` and contains a workaround for the way Swashbuckle handles generic types. Please see the comment in the code for explanation of why it was added and how it works.

P.S. I tried to use the `WebHelper.UrlEncode(...)` method instead of replaces - it seems to be a more correct solution. But it causes the Swagger Editor to fail to resolve any generic type references. So, `UrlEncode` probably can't be used here.

### DIFF
--- a/VirtoCommerce.Platform.Web/Swagger/SwaggerConfig.cs
+++ b/VirtoCommerce.Platform.Web/Swagger/SwaggerConfig.cs
@@ -1,7 +1,8 @@
-﻿using System;
+using System;
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
+using System.Net;
 using System.Net.Http;
 using System.Reflection;
 using System.Web.Hosting;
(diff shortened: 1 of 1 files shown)


avas:
Some more additions:
1. I've checked the Swagger validation badge for the demo platform instance - it shows that the Swagger schema for this instance is valid:
    Swagger validation result
    However, it performs the validation [differently than the Swagger Editor](https://github.com/swagger-api/validator-badge/issues/117) (and probably differently from offline validation utilities like `swagger-cli`).
2. I've also tried to build the AutoRest client for modified platform and check generic types, and it seems like AutoRest does not use generic types at all. For example, the `CustomerModuleController.Search` method returns an `GenericSearchResult<Member>`, but in AutoRest-generated client it returns plain `GenericSearchResultMember`. This behavior was like this even before my modifications, and this PR doesn't affect the AutoRest-generated code.
image

PR #1475, opened 2018-11-16
Account API blade fixed

priboltik:
### Problem
JS error in console while new API Account blade opening

### Solution
Need to remove all errors in console.

### Proposed of changes
Added additional condition for blade.currentEntity object.


tatarincev:
Released https://github.com/VirtoCommerce/vc-platform/releases/v2.13.38

Issue #1476, opened 2018-11-16
Replace cookie-based authentication to token-based for manager SPA


tatarincev:
Technical details:
The platform already has the support of OAuth Token based authentication http://bitoftech.net/2014/06/01/token-based-authentication-asp-net-web-api-2-owin-asp-net-identity/
Need to perform the following sub-tasks to completely switch the manager SPA to the token based authentication. 

- Implement OAuth  Refresh tokens  for platform application
http://bitoftech.net/2014/07/16/enable-oauth-refresh-tokens-angularjs-app-using-asp-net-web-api-2-owin/
- Need to duplicate logic that presents in in the [SecurityController.Login](https://github.com/VirtoCommerce/vc-platform/blob/master/VirtoCommerce.Platform.Web/Controllers/Api/SecurityController.cs#L67) 
 in the [ApplicationOAuthProvider](https://github.com/VirtoCommerce/vc-platform/blob/master/VirtoCommerce.Platform.Web/ApplicationOAuthProvider.cs#L12) that will be used for authentication
- Extend the current [auth service](https://github.com/VirtoCommerce/vc-platform/blob/master/VirtoCommerce.Platform.Web/Scripts/app/security/services/auth.js) to use OAuth **implicit Grant**  flow based on token authorization with refresh token instead of cookies. Need to store received auth tokens in the browser local storage. Also need to use `refresh` flow for prolongation the session of  authentificated user. 
_Example ready OAuth clients for AngularJS:_ 
https://github.com/sahat/satellizer
https://github.com/oauthjs/angular-oauth2

tatarincev:
Released https://github.com/VirtoCommerce/vc-platform/releases/tag/v2.13.39

PR #1480, opened 2018-11-28
1476: migrating VC Platform Manager to bearer token authentication

avas:
This PR migrates the VC Platform Manager from cookie-based authentication to bearer token-based authentication (#1476). Changes made to the platform to make it happen:
* Added the `RefreshTokenEntity` to store refresh tokens in the database. It can be accessed using the `IPlatformRepository.RefreshTokens` property.
* Added the `IRefreshTokenService` (implemented by `RefreshTokenServiceImpl`) which can be used to retrieve, store and delete refresh tokens. It's main purpose is to decouple the `RefreshTokenProvider` from the `IPlatformRepository`.
* Added the `RefreshTokenProvider` which is used by the OWIN OAuth server infrastructure to work with refresh tokens. It is registered and used in the `OwinConfig`.
* `platformWebApp.authService`: `login()` method does not call the `SecurityController.Login()` method anymore - it requests the access token instead. This way, the cookie-based authentication won't be used for regular sign-ins.
* `platformWebApp.httpErrorInterceptor` now controls access tokens and adds headers required for authentications. It adds required headers to requests to platform API and requests new access tokens if it detects token expiration. To access the authentication data, it uses the `platformWebApp.authDataStorage` service which internally uses the `angular-local-storage` module.

Authentication using Azure AD also works, but requires the cookie-based authentication to be enabled.


VirtoCommerce.Platform.Web/Scripts/app/security/services/auth.js
```diff
@@ -10,21 +10,68 @@ angular.module('platformWebApp')
...
+                authContext.logout();
+                return $q.reject(error);
             });
     };
-    
+
+    authContext.refreshToken = function() {
+        var authData = authDataStorage.getStoredData();
+        if (authData) {
+            var data = 'grant_type=refresh_token&refresh_token=' + authData.refreshToken;
```
artem-dudarev: Token should be URL-encoded before adding to data

VirtoCommerce.Platform.Data.Security/Services/RefreshTokenServiceImpl.cs
```diff
@@ -0,0 +1,73 @@
...
+using System.Linq;
+using System.Threading.Tasks;
+using VirtoCommerce.Platform.Core.Common;
+using VirtoCommerce.Platform.Core.Security;
+using VirtoCommerce.Platform.Data.Model;
+using VirtoCommerce.Platform.Data.Repositories;
+
+namespace VirtoCommerce.Platform.Data.Security.Services
+{
+    public class RefreshTokenServiceImpl : IRefreshTokenService
```
artem-dudarev: Don't use `Impl` suffix. There is no such word.

VirtoCommerce.Platform.Web/Scripts/app/security/services/auth.js
```diff
@@ -10,21 +10,68 @@ angular.module('platformWebApp')
...
             },
             function (error) { });
     };
 
     authContext.login = function (email, password, remember) {
-        return $http.post(serviceBase + 'login/', { userName: email, password: password, rememberMe: remember }).then(
-            function (results) {
-                changeAuth(results.data);
-                return authContext.isAuthenticated;
+        var requestData = 'grant_type=password&username=' + email + '&password=' + password;
```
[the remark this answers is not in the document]
avas: Sorry, I forgot about it. Should be fixed now.

VirtoCommerce.Platform.Data.Security/Services/RefreshTokenServiceImpl.cs
avas: Got it, renamed to `RefreshTokenService`.

avas:
I have added the helper cookie that would help to authorize non-AJAX requests with no bearer token. It would help in some cases that didn't work with previous implementation, such as:
* Downloading platform export result file;
* Accessing assets;
* Using Hangfire dashboard.

How it works:
1. `OwinConfig.cs`: if the bearer token authentication is enabled, we register yet another cookie middleware (`AuthenticationType` = `BearerTokenHelperCookie`).
2. `ApplicationOAuthProvider`: if user credentials are valid, we issue a helper cookie in addition to the authentication token. This way, the user will be able to authenticate even without providing a security token.
3. When the user makes a request to the platform, they are challenged against the helper cookie *and* the authentication token:
    * If the user provided authentication token only, they will be authenticated using the bearer token middleware and granted the identity with `AuthenticationType = Bearer`.
    * If the user provided helper cookie only, they will be authenticated with helper cookie middleware and granted the identity with `AuthenticationType = BearerTokenHelperCookie`.
    * If both cookie and token were provided, the user will be authenticated with *last* middleware in the OWIN pipeline that was able to successfully authenticate the user. In this case, the bearer token authentication is added *after* the helper cookie middleware, so the user will be authenticated with bearer token authentication middleware and granted the identity with `AuthenticationType = Bearer`.
4. `CheckPermissionAttribute`: if the user is authenticated using the helper cookie only (`identity.AuthenticationType == BearerTokenHelperCookie`), their permission will be limited to the permissions listed in the `VirtoCommerce:Authentication:BearerTokens.PermissionsToCheck` configuration property. The attribute accesses properties using the `ICheckPermissionAttributeSettings` interface implemented by the `CheckPermissionAttributeSettingsAdapter` class (which uses the `AuthenticationOptions` class as the source for settings).

This way, the export file download and the Hangfire dashboard work successfully (I checked it). Assets file access should also work now. All requests in Swagger UI that require only the `security:call_api` permission also work, but requests that require elevated permissions (e.g. `GET /api/platform/security/users/{userName}`) will fail with error 401.

avas:
After discussing these changes with @tatarincev, we decided to make some more changes:
* Remove separate cookies middleware for helper cookie validation - instead, we should use the main cookies;
* To store whitelist of permissions granted to user, we should use the additional grant `LimitedPermissions`. If it is defined for the user identity, `CheckPermissionAttribute` should grant only  permissions listed there; if it is not defined, `CheckPermissionAttribute` should not limit user permissions.

I checked this approach, and it appears to work. So, the only thing left to settle is the bearer token authentication in Swagger UI.

Issue #1467, opened 2018-11-09
Allow to  change shortcut icon for platform application


vturovets:
**Problem:**
Need to add possibility  to change shortcut icon  for the platform application

tatarincev:
Technical details:
The platform has a special setting that allows to do some UI customization. 
![image](https://user-images.githubusercontent.com/7566324/48331069-9c158780-e657-11e8-9954-a45ecbe55d69.png)
We need to extend this setting with the new parameter with name `favicon` that can contains path to the icon that is used as shortcut icon for application.
Also need to somehow to read `favicon` setting value and use it in the `_Layout.cshtm` or 
 use the default if this  setting not set. https://github.com/VirtoCommerce/vc-platform/blob/master/VirtoCommerce.Platform.Web/Views/Shared/_Layout.cshtml#L8

tatarincev:
The default favicon path doesn't work
![image](https://user-images.githubusercontent.com/7566324/49726898-b4d58500-fc77-11e8-8413-d0bf9f08a805.png)

yecli:
Changed default favicon path: "~/favicon.ico" => "favicon.ico".

tatarincev:
Released https://github.com/VirtoCommerce/vc-platform/releases/tag/v2.13.39

PR #1483, opened 2018-12-03
#1467 Allow to change shortcut icon for platform application

yecli:
#1467 Allow to change shortcut icon for platform application 

Added "favicon" setting into "VirtoCommerce.Platform.UI.Customization", value is passed to Layout through ViewBag.

### DIFF
--- a/VirtoCommerce.Platform.Web/Controllers/HomeController.cs
+++ b/VirtoCommerce.Platform.Web/Controllers/HomeController.cs
@@ -50,7 +50,7 @@ namespace VirtoCommerce.Platform.Web.Controllers
                 }
             }
 
-            ViewBag.FavIcon = GetFavIcon();
+            var favIcon = GetFavIcon();
 
             return View(new PlatformSetting
             {
@@ -58,6 +58,7 @@ namespace VirtoCommerce.Platform.Web.Controllers
                 DemoCredentials = new MvcHtmlString(demoCredentials ?? "''"),
                 DemoResetTime = new MvcHtmlString(resetTimeStr ?? "''"),
                 License = new MvcHtmlString(licenseString),
+                FavIcon = new MvcHtmlString(favIcon ?? "~/favicon.ico"),
             });
         }
 
--- a/VirtoCommerce.Platform.Web/Controllers/HomeController.cs
+++ b/VirtoCommerce.Platform.Web/Controllers/HomeController.cs
@@ -3,8 +3,10 @@ using System.Reflection;
 using System.Web.Hosting;
 using System.Web.Mvc;
 using Newtonsoft.Json;
+using Newtonsoft.Json.Linq;
 using Newtonsoft.Json.Serialization;
 using VirtoCommerce.Platform.Core.Common;
+using VirtoCommerce.Platform.Core.Settings;
 using VirtoCommerce.Platform.Web.Licensing;
 using VirtoCommerce.Platform.Web.Model;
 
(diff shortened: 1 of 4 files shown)


VirtoCommerce.Platform.Web/Controllers/HomeController.cs
```diff
@@ -41,6 +50,8 @@ public ActionResult Index()
                 }
             }
 
+            ViewBag.FavIcon = GetFavIcon();
```
tatarincev: maybe will be better extend `PlatformSetting` with new property `FavIcon` instead of define the new property in ViewBag? It will be more `by current design`, thoughts?

yecli: Is seems strange to me to use Model in layout (as it could be used in many pages with different models - and we expect all of them to have some property).
In our case - layout is used in one page, so we know what model we could use.
In general i think ViewBag is more appropriate for this as it is dynamic collection initially, and model is strongly typed.
In case we are not going to create additional views with this layout - we could extend PlatformSetting.
What do you think?

yecli: Or maybe we could use this approach:
https://stackoverflow.com/a/14882183
We will have strongly typed object then.

tatarincev: You right, it is the absolutely not `by right way` solution  to access in Layout any of concrete view models, but in your case we have a SPA application that has  only one view and this works. Regarding of   question of using ViewBug or strongly typed ViewModel in layout the both cases are incorrect due that fact described above. Why I'm leaning towards the extending `PlatformSetting` type with new property? Because  this approach already exists and works, and these changes will be consistent with the current design.

yecli: Ok, agreed, will extend PlatformSetting to be consistent with current approach.

yecli: Done

VirtoCommerce.Platform.Web/Controllers/HomeController.cs
```diff
@@ -41,12 +50,15 @@ public ActionResult Index()
                 }
             }
 
+            var favIcon = GetFavIcon();
```
tatarincev: Seems it is redundant variable,  GetFavIcon() can be called in place.

VirtoCommerce.Platform.Web/Controllers/HomeController.cs
```diff
@@ -68,5 +80,17 @@ private static License LoadLicense()
...
         }
+
+        private string GetFavIcon()
+        {
+            string result = null;
+            var uiSettings = _settingsManager.GetSettingByName("VirtoCommerce.Platform.UI.Customization");
+            if (uiSettings != null)
+            {
+                var jObject = JObject.Parse(uiSettings.Value);
+                result = (string)jObject.SelectToken("favicon");
```
tatarincev: Need to take into account what there is may be platform installations  with already exist stored settings which may not   contain the new  `favicon` property.   So then this lines mustn't cause an exception.

VirtoCommerce.Platform.Web/Controllers/HomeController.cs
yecli: Done

VirtoCommerce.Platform.Web/Controllers/HomeController.cs
yecli: Handled possible JObject.Parse exceptions and made explicit expression suppress on jObject.SelectToken.

PR #1497, opened 2018-12-12
Features 1330 client error handling

yecli:
- Logout user if he deletes himself
- Added client error handling for import/export
image

### DIFF
--- a/VirtoCommerce.Platform.Web/Controllers/Api/SecurityController.cs
+++ b/VirtoCommerce.Platform.Web/Controllers/Api/SecurityController.cs
@@ -551,6 +551,11 @@ namespace VirtoCommerce.Platform.Web.Controllers.Api
         [CheckPermission(Permission = PredefinedPermissions.SecurityDelete)]
         public async Task<IHttpActionResult> DeleteAsync([FromUri] string[] names)
         {
+            if (names.Contains(User.Identity.Name))
+            {
+                throw new HttpException((int)HttpStatusCode.InternalServerError, "You cannot delete your own account.");
+            }
+
             EnsureUserIsEditable(names);
 
             await _securityService.DeleteAsync(names);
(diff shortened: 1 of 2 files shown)


VirtoCommerce.Platform.Web/Scripts/app/security/blades/account-list.js
```diff
@@ -46,7 +46,12 @@ function ($scope, accounts, dialogService, uiGridHelper, bladeNavigationService,
                 if (remove) {
                     bladeNavigationService.closeChildrenBlades(blade, function () {
                         var itemIds = _.pluck(selection, 'userName');
-                        accounts.remove({ names: itemIds }, blade.refresh);
+                        accounts.remove({ names: itemIds }, function () {
```
tatarincev: I don't think that is good idea to allow user remove self account from system, could we prevent this action?

VirtoCommerce.Platform.Web/Scripts/app/exportImport/blades/export-main.js
```diff
@@ -10,6 +10,8 @@
         exportImportResourse.getNewExportManifest(function (data) {
             $scope.exportRequest.exportManifest = data;
             blade.isLoading = false;
+        }, function (error) {
```
tatarincev: Could you check why the common platform error handling doesn't work for this actions?
https://github.com/VirtoCommerce/vc-platform/blob/master/VirtoCommerce.Platform.Web/Scripts/app/app.js#L59

VirtoCommerce.Platform.Web/Scripts/app/security/blades/account-list.js
yecli: Yes, we could do this. But in many systems removing self account is a legal action (like Github). Maybe better to add confirmation that user is trying to delete himself (or add that info to initial confirmation on Delete action)?

tatarincev: Yes, confirmation will be a good choice.

yecli: Decided to restrict possibility of self deletion on server side.

VirtoCommerce.Platform.Web/Scripts/app/exportImport/blades/export-main.js
yecli: It worked, in fact, so reverted redundant handlers. Also changed error handler for FileUploader to produce error with non empty details.

PR #1503, opened 2018-12-26
CMS Integration

basilkot:
Add using CORS by parameters in web.config.
* add paramters with prefix `cors:`, which can be disabled when one of them is blank
* add `SameSite` paramter to cookies to allow cross-domain access to API

All of these changes are not needed when CMS designer is run in the same domain as Platform


tatarincev:
Implemented with the different way by this commit
https://github.com/VirtoCommerce/vc-platform/commit/5e7a7404ff8417337ce2e24214318c47ec51a7f4

Issue #1507, opened 2019-01-09
Simplify platform configuration for working with Redis cache server


tatarincev:
**Problem:**
Need to have possibility switch the platform cache configuration  to work with Redis server by change only one application setting (probably add connection Redis string). This is now achieved by manually changing the configuration described in this article.
https://virtocommerce.com/docs/vc2devguide/deployment/platform-scaleout

tatarincev:
Released https://github.com/VirtoCommerce/vc-platform/releases/v2.13.41

PR #1508, opened 2019-01-10
#1507 Added appSettings parameter to control using of VC cache instances

Andrew-Orlov:
### Problem
Need to have possibility switch the platform cache configuration to work with Redis server by change only one application setting (probably add connection Redis string).

### Solution
Added a new setting in appSettings to control using of scaled up VC cache instances

### Proposed of changes
- Added a new parameter `VirtoCommerce:Cache:ScaleUpEnabled` to appSettings
- Added new cache manager to switch
- Added checks to decide which cache manager should be chosen (default mem cache or scaled up within Redis) or create a default mem cache manager if nothing was found

### DIFF
--- a/VirtoCommerce.Platform.Web/Startup.cs
+++ b/VirtoCommerce.Platform.Web/Startup.cs
@@ -419,13 +419,13 @@ namespace VirtoCommerce.Platform.Web
                 var defaultCacheManager = cacheManagerSection.CacheManagers.FirstOrDefault(p => p.Name.EqualsInvariant("defaultPlatformCache"));
                 if (defaultCacheManager != null)
                 {
-                    configuration = ConfigurationBuilder.LoadConfiguration("defaultPlatformCache");
+                    configuration = ConfigurationBuilder.LoadConfiguration(defaultCacheManager.Name);
                 }
 
                 var redisCacheManager = cacheManagerSection.CacheManagers.FirstOrDefault(p => p.Name.EqualsInvariant("redisPlatformCache"));
                 if (cacheScaleUpEnabled && redisCacheManager != null)
                 {
-                    configuration = ConfigurationBuilder.LoadConfiguration("redisPlatformCache");
+                    configuration = ConfigurationBuilder.LoadConfiguration(redisCacheManager.Name);
                 }
 
                 if (configuration != null)
--- a/VirtoCommerce.Platform.Web/Startup.cs
+++ b/VirtoCommerce.Platform.Web/Startup.cs
@@ -406,7 +406,7 @@ namespace VirtoCommerce.Platform.Web
             app.SanitizeThreadCulture();
             ICacheManager<object> cacheManager = null;
 
-            var redisConnectionString = ConfigurationManager.ConnectionStrings["RedisConnectionString"];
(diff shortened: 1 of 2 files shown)


VirtoCommerce.Platform.Web/Startup.cs
```diff
@@ -406,12 +406,27 @@ private static void InitializePlatform(IAppBuilder app, IUnityContainer containe
             app.SanitizeThreadCulture();
             ICacheManager<object> cacheManager = null;
 
+            // Try to find app setting which controls using of scaled up VC cache instances
+            var cacheScaleUpSetting = ConfigurationManager.AppSettings["VirtoCommerce:Cache:ScaleUpEnabled"];
```
tatarincev: I think we should replace of using the special flag to read the redisConnection from web.config   <connectionStrings> section.  that can be taken by follow call in startup.cs.
CacheManager 
```C#
ConfigurationHelper.GetConnectionStringValue("redisConnectionString")
```
 Need somehow find out solution how to use this connection string instead of special configuration section. Because it is impossible to manage this custom section in Azure portal
```XML
<cacheManager.Redis>
        <connections>
...
```

VirtoCommerce.Platform.Web/Web.config
```diff
@@ -112,22 +113,43 @@
...
+            <cache name="redisPlatformCache" enableStatistics="true" backplaneName="redisConnectionString" backplaneType="VirtoCommerce.Platform.Web.Cache.RedisCacheBackplane2, VirtoCommerce.Platform.Web">
                 <handle name="memCacheHandle" ref="memCacheHandle" expirationMode="Sliding" timeout="10m" />
+                <handle name="redisConnectionString" ref="redisHandle" isBackplaneSource="true" />
             </cache>
         </managers>
         <cacheHandles>
+            <handleDef id="redisHandle" type="VirtoCommerce.Platform.Web.Cache.RedisCacheHandle2`1, VirtoCommerce.Platform.Web" />
             <handleDef id="memCacheHandle" type="CacheManager.SystemRuntimeCaching.MemoryCacheHandle`1, CacheManager.SystemRuntimeCaching" />
         </cacheHandles>
+
```
tatarincev: This comments are redundant

VirtoCommerce.Platform.Web/Web.config
```diff
@@ -112,22 +113,43 @@
...
+        </cacheHandles>-->
     </cacheManager>
     <system.runtime.caching>
         <memoryCache>
             <namedCaches>
                 <!--For some cases, this setting physicalMemoryLimitPercentage = "memory usage threshold in percents",
                 leads to a complete stop of the cache working even after the current physical memory usage returns less that set's threshold--> 
-                <add name="memCacheHandle" />
+                <!--<add name="memCacheHandle" />-->
+                <add name="memCacheHandle" physicalMemoryLimitPercentage="80" pollingInterval="00:00:30" />
```
tatarincev: this line also need to be removed

Andrew-Orlov: Should it be removed actually? Are these parameters useless?

PR #1519, opened 2019-02-08
BE: IQueryableExtensions.OrderBy now is working with properties of derived types.

blazemachine:
Effective type (type of elements in list) was added to code.

### DIFF
--- a/VirtoCommerce.Platform.Core/Common/IQueryableExtensions.cs
+++ b/VirtoCommerce.Platform.Core/Common/IQueryableExtensions.cs
@@ -1,40 +1,39 @@
 using System;
-using System.Collections.Generic;
 using System.Linq;
-using System.Text;
 using System.Linq.Expressions;
 using System.Reflection;
-using System.Collections;
 
 namespace VirtoCommerce.Platform.Core.Common
 {
-	public static class IQueryableExtensions
-	{
-		public static IOrderedQueryable<T> OrderBy<T>(this IQueryable<T> source, string property)
-		{
-			return ApplyOrder<T>(source, property, "OrderBy");
-		}
-		public static IOrderedQueryable<T> OrderByDescending<T>(this IQueryable<T> source, string property)
-		{
-			return ApplyOrder<T>(source, property, "OrderByDescending");
-		}
-		public static IOrderedQueryable<T> ThenBy<T>(this IOrderedQueryable<T> source, string property)
-		{
-			return ApplyOrder<T>(source, property, "ThenBy");
-		}
-		public static IOrderedQueryable<T> ThenByDescending<T>(this IOrderedQueryable<T> source, string property)
-		{
-			return ApplyOrder<T>(source, property, "ThenByDescending");
-		}
+    public static class IQueryableExtensions
+    {
+        public static IOrderedQueryable<T> OrderBy<T>(this IQueryable<T> source, string property)
+        {
+            return ApplyOrder(source, property, "OrderBy");
+        }
 
-      
+        public static IOrderedQueryable<T> OrderByDescending<T>(this IQueryable<T> source, string property)
+        {
+            return ApplyOrder(source, property, "OrderByDescending");
+        }
+
+        public static IOrderedQueryable<T> ThenBy<T>(this IOrderedQueryable<T> source, string property)
+        {
+            return ApplyOrder(source, property, "ThenBy");
+        }
+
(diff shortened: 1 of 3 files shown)


VirtoCommerce.Platform.Core/Common/IQueryableExtensions.cs
```diff
@@ -69,39 +69,40 @@ public static IOrderedQueryable<T> ThenBySortInfos<T>(this IOrderedQueryable<T>
...
-				throw new ArgumentNullException("property");
+            if (property == null)
+                throw new ArgumentNullException("property");
 
-			string[] props = property.Split('.');
-			Type type = typeof(T);
-			ParameterExpression arg = Expression.Parameter(type, "x");
-			Expression expr = arg;
+            string[] props = property.Split('.');
+            var effectiveType = source.FirstOrDefault().GetType();
```
artem-dudarev: FirstOrDefault() will lead to redundant and unexpected request to the database.

yecli: Really, it is non-acceptable solution. Cannot find a way to get specific element runtime type besides getting element. [IQueryable.ElementType](https://docs.microsoft.com/en-US/dotnet/api/system.linq.iqueryable.elementtype?view=netframework-4.7.2) does not help.  [System.Linq.Dynamic](https://github.com/kahanu/System.Linq.Dynamic) cannot handle such situations too.
So we should admit that we cannot implement sorting using derived class new properties in this extension without immediate execution.
The only viable option I see is passing type as propertyName, in string. 
So need to decide if additional extension like .OrderBy(string propertyName, string elementTypeName) could be useful for us.

yecli: Decided to use type registered in AbstractTypeFactory for base type.

VirtoCommerce.Platform.Core/Common/IQueryableExtensions.cs
```diff
@@ -68,40 +67,42 @@ public static IOrderedQueryable<T> ThenBySortInfos<T>(this IOrderedQueryable<T>
...
         public static IOrderedQueryable<T> ApplyOrder<T>(IQueryable<T> source, string property, string methodName)
-		{
-			if (property == null)
-				throw new ArgumentNullException("property");
+        {
+            if (property == null)
+                throw new ArgumentNullException(nameof(property));
+
+            var props = property.Split('.');
+            var registeredType = AbstractTypeFactory<T>.AllTypeInfos.Select(x => x.Type).FirstOrDefault();
```
yecli: Better to use AbstractTypeFactory<T>.TryCreateInstance(), as it consider inheritance.

VirtoCommerce.Platform.Tests/Common/QueryableExtensionsTests.cs
```diff
@@ -0,0 +1,87 @@
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using VirtoCommerce.Platform.Core.Common;
+using Xunit;
+
+namespace VirtoCommerce.Platform.Tests.Common
+{
+    [CLSCompliant(false)]
+    public class QueryableExtensionsTests
```
yecli: Please add     
[Trait("Category", "CI")] 
to include these tests into CI process

anivirtoway: fix

VirtoCommerce.Platform.Core/Common/IQueryableExtensions.cs
anivirtoway: fix

VirtoCommerce.Platform.Core/Common/IQueryableExtensions.cs
yecli: @artem-dudarev Could you check new implementation?

anivirtoway:
To use EF you cannot use the expression "Expression expr = Expression.Convert (arg, effectiveType);" more details can be found here https://entityframework.net/knowledge-base/47425013/linq-to-entities-only-supports-casting-edm-primitive-or-enumeration-types--unique-new-post-
Unfortunately I did not find an alternative

tatarincev:
Have been moved to On Hold  because the proposed solution should not allow cases with sorting collections containing objects of different types.

PR #1524, opened 2019-02-25
Make sevaral classes cloneable.

vanzegbroecke:
We want to be able to clone Organizations and Contacts. Use case: MemberChangedEvent handlers which update the member. We can't update the newEntity because otherwise other handlers will not get the original newEntity. To do so , we need several classes to be cloneable.

Needed for: https://github.com/VirtoCommerce/vc-module-core/pull/111


vanzegbroecke:
> Could you make all these Clone methods as virtual?

Certainly, please see commit a6f0a4b6aa64dc6d73ffe70a830dbb57706cdaa4.

yecli:
Released in [v2.13.42](https://github.com/VirtoCommerce/vc-platform/releases/tag/v2.13.42)

Issue #1526, opened 2019-02-26
Add Twilio and ASPSMS SMS providers support + new notification type


yecli:
1. Need to add ResetPasswordSmsNotification with string Token for sending verification code on password reset.
2. Need to support two SMS providers: Twilio and ASPSMS, configurable through web.config.

yecli:
Demo:
![jwtrnjunoa](https://user-images.githubusercontent.com/15610352/53486136-bc077480-3aa9-11e9-9fc9-8d6e3f8a400f.gif)

yecli:
Released in [v2.13.42](https://github.com/VirtoCommerce/vc-platform/releases/tag/v2.13.42)

PR #1527, opened 2019-02-27
#1526 Add Twilio and ASPSMS SMS providers support + new notification type

yecli:
Added SMS providers and new notification template


VirtoCommerce.Platform.Web/Startup.cs
```diff
@@ -529,6 +541,62 @@ private static void InitializePlatform(IAppBuilder app, IUnityContainer containe
                         }
                     },
                     new ModuleSettingsGroup
+                    {
```
tatarincev: We don't need a create platform settings that can be changed in run-time, let's restricting only settings that being  reading from web.config

yecli: Done

VirtoCommerce.Platform.Data.Notifications/SMSGatewayOptions.cs
```diff
@@ -0,0 +1,22 @@
+namespace VirtoCommerce.Platform.Data.Notifications
+{
+    public class SmsGatewayOptions
```
tatarincev: I think that make sense create the two different Options for each provider type and inject or pass into ctor of concrete provider only  settings that it understand.
Factory for this options you can implement in startup as well

yecli: Committed changes regarding this. If it is too complex - could add two different classes and option sets.

PR #1547, opened 2019-03-29
modification checkPermission in authService

artmakarov:
Added verification: if we have no scope means we have permission

### DIFF
--- a/VirtoCommerce.Platform.Web/Scripts/app/security/services/auth.js
+++ b/VirtoCommerce.Platform.Web/Scripts/app/security/services/auth.js
@@ -99,6 +99,20 @@ angular.module('platformWebApp')
         changeAuth({});
     };
 
+    /**
+     * if we are checking
+     * authService.checkPermission('order:read', 'limitResponseGroup:WithPrices');
+     *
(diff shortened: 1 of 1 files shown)


VirtoCommerce.Platform.Web/Scripts/app/security/services/auth.js
```diff
@@ -119,7 +119,7 @@ angular.module('platformWebApp')
 
                         // if we have no scope means we have permission
                         var retVal = !_.some(authContext.permissions, function (item) {
-                            return item === scopeName || item.indexOf(scopeName + ':') > -1;
+                            return item === scopeName || !item.indexOf(scopeName + ':');
```
yecli: As i understand, thus we want to check if string starts with another. I think 
`item.indexOf(scopeName + ':') === 0`
is better to understand than 
`!item.indexOf(scopeName + ':')` 
in that case.

artmakarov: fixed

yecli:
WithPrices was implemented as permission, so these changes are not needed.

PR #1550, opened 2019-04-03
Paged search ordering revision + optimization

AliveMen:
added method OrderBySortInfosWithId


yecli:
Decided not to force update platform for this extension method. Common solution would be Linq .ThenBy("Id").

PR #1572, opened 2019-04-25
Added disabled boolean checkbox on readonly property

pushnitsa:
Issue comes when metafields have readonly boolean property. Current PR will fix this.


yecli:
Released in [v2.13.45](https://github.com/VirtoCommerce/vc-platform/releases/tag/v2.13.45).

PR #1587, opened 2019-06-03
fix settings blade sorting

anivirtoway:
before - no settings ordering

after - sort settings by name


anivirtoway:
> any changes to allPackages.js should be done separately. Can exclude this change?
done

Issue #1592, opened 2019-06-11
Can't add email to copy or blind copy when use emailnotifications


AkulovOleg:
**Describe the bug**
When i add emails to bcc or cc properties of EmailNotification object and call ScheduleSendNotification method i don't receive email to that addresses. 

BCC and CC don't save to database, so when job take notifications from database, BCC and CC properties are empty.
https://github.com/VirtoCommerce/vc-platform/blob/853d66a976c47bf8d7b2c5222c6acf2bfc720915/VirtoCommerce.Platform.Data/Notifications/NotificationConverter.cs#L10

 - Platform version: 2.13.47

tatarincev:
Released
https://github.com/VirtoCommerce/vc-platform/releases/tag/v2.13.48

PR #1603, opened 2019-07-10
#1592 Notification add cc,bcc fields to database

kostyrin:


VirtoCommerce.Platform.Core/Notifications/EmailNotification.cs
```diff
@@ -13,6 +9,6 @@ protected EmailNotification(IEmailNotificationSendingGateway gateway) : base(gat
         }
 
         public IList<string> CC { get; set; } = new List<string>();
-        public IList<string> Bcc { get; set; } = new List<string>();
+        public IList<string> BCC { get; set; } = new List<string>();
```
artem-dudarev: Undo.
By convention, two-letter acronyms use all uppercase letters, and acronyms of three or more characters use Pascal casing.
https://docs.microsoft.com/en-us/visualstudio/code-quality/ca1709-identifiers-should-be-cased-correctly?view=vs-2019

Issue #1590, opened 2019-06-04
User password changing detaches this users roles from him. 


trueboroda:
I`ve changed the password of some user via platform security tools. Then i see that user doesn't have his roles anymore.

tatarincev:
Released 
https://github.com/VirtoCommerce/vc-platform/releases/tag/v2.13.48

PR #1604, opened 2019-07-10
#1590 don't detaches roles

kostyrin:


kostyrin:
> Checked implementation - I think Roles and ApiAccounts should not be erased even if we get user with `UserDetails.Reduced` detail group - protection with IsNullCollection check exists. So imho the problem is that when we are using Reduced group we are getting empty Roles and ApiAccounts arrays instead of null. So I'd better put clearing of these properties in `SecurityService.GetUserExtendedAsync` for truncated corresponding groups (assuming I get correctly that Reduce should not contain Roles and ApiAccounts).
> image
> I think we need to ask @tatarincev on this.

ok, I'll try to fix it

PR #1619, opened 2019-08-06
Support VS 2019 in Compress-Module script

### LINKED ISSUE #1618: Compress-Module doesn't work properly in Visual Studio 2019
Compress-Module cannot find MSBuild.exe that is installed with VS 2019 and falls back to MSBuild.exe in the Windows folder, which cannot compile C# 7 code.

https://github.com/VirtoCommerce/vc-platform/blob/master/VirtoCommerce.Module/tools/VirtoCommerce.Module.psm1#L23

Need to update VirtoCommerce.Module.psm1 to support new path for MSBuild.exe in VS 2019, which is 'MSBuild\\**Current**\Bin\MSBuild.exe'.

artem-dudarev:
Fixes #1618

### DIFF
--- a/VirtoCommerce.Module/tools/VirtoCommerce.Module.psm1
+++ b/VirtoCommerce.Module/tools/VirtoCommerce.Module.psm1
@@ -1,4 +1,4 @@
-﻿function Compress-Module
+function Compress-Module
 {
     Param(
         [string] $ProjectName,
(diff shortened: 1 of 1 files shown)


yecli:
Released in https://github.com/VirtoCommerce/vc-platform/releases/tag/v2.13.50

PR #1621, opened 2019-08-19
FE: Additional HTTP code 400 for incorrect login or password is checking now.

blazemachine:


akak1977:
Released: https://github.com/VirtoCommerce/vc-platform/releases/tag/v2.13.54

PR #1622, opened 2019-08-23
VP-310: Added "export:download" permission to "LimitedCookiePermissions" for possibility to download export file;

yecli:
Added "export:download" permission to "LimitedCookiePermissions" for possibility to download export file;


yecli:
Not needed, platform:export was used.

PR #1629, opened 2019-09-03
FE, BE: Localization for dynamic properties descriptions.

blazemachine:


VirtoCommerce.Platform.Data/Repositories/Migrations/201908221604169_DynamicPropertyNameAddDescription.cs
```diff
@@ -0,0 +1,40 @@
...
+    using System;
+    using System.Data.Entity.Migrations;
+    
+    public partial class DynamicPropertyNameAddDescription : DbMigration
+    {
+        public override void Up()
+        {
+            AddColumn("dbo.PlatformDynamicPropertyName", "Description", c => c.String(maxLength: 256));
+
+            Sql(@"DECLARE @cursor CURSOR
```
tatarincev: I think a better solution will be using simple SQL Update statement instead of use Cursor and Fetch here

VirtoCommerce.Platform.Web/Scripts/app/dynamicProperties/blades/dynamicProperty-detail.tpl.html
```diff
@@ -102,6 +102,14 @@
                         <input ng-model="data.name" placeholder="{{ 'platform.blades.dynamicProperty-detail.placeholders.display-name' | translate }}" />
                     </div>
                 </div>
+                <div class="form-group">
```
tatarincev: More reasonable will be put input tag for description near with `DisplayName` in line 102, isn't it?

VirtoCommerce.Platform.Web/Scripts/app/dynamicProperties/blades/dynamicProperty-detail.tpl.html
```diff
@@ -88,7 +88,7 @@
                         <input class="form-input" num-type="integer" type="number" ng-min="0" min="0" ng-model="blade.currentEntity.displayOrder" placeholder="{{ 'platform.blades.dynamicProperty-detail.placeholders.display-order' | translate}}" />
                     </div>
                 </div>
-                <div class="form-group">
+                <div class="form-group" hidden>
```
tatarincev: Need just remove this input if it no longer needed

VirtoCommerce.Platform.Web/Scripts/app/dynamicProperties/dynamicProperties.js
```diff
@@ -56,4 +56,13 @@
 
         return retVal;
     }
-});
\ No newline at end of file
+})
```
tatarincev: I don't understand why we need this code, we need the display Description for a dynamic property as well as DisplayName, I don see any differences, why we required this code then?

VirtoCommerce.Platform.Data/Repositories/Migrations/201908221604169_DynamicPropertyNameAddDescription.cs
blazemachine: Replaced.

VirtoCommerce.Platform.Web/Scripts/app/dynamicProperties/blades/dynamicProperty-detail.tpl.html
blazemachine: This field was needed because at update of field earlier old description could be lost. Now this behaviour is not observing. This field will be remove.

VirtoCommerce.Platform.Web/Scripts/app/dynamicProperties/blades/dynamicProperty-detail.tpl.html
blazemachine: Here will be a many excess labels tags: name and description for every language. Even we are will be grouping them by language. It will be looking bad in other cases. This is may to stay so or need to reconstruct design of this fully.

VirtoCommerce.Platform.Web/Scripts/app/dynamicProperties/dynamicProperties.js
blazemachine: This filter is strange. He has a name that speak about a localization of name but factually besides localization he build a formatted string. We need a various logic of view for description and name (in this case name has parentheses and description are not). At logic in one filter we will have a hard-code for description and name and other properties. A new filter don't do this. And old filter will work on.

yecli:
Clothing as v2 in LTS

PR #1734, opened 2019-12-02
#vp-975 Swagger filter by tag

kostyrin:


akak1977:
Is it possible to search case-insensitively?

PR #1737, opened 2019-12-04
the platform's version info details downloading ability 

trueboroda:


VirtoCommerce.Platform.Web/Converters/Diagnostics/InstalledModuleInfoConverter.cs
```diff
@@ -0,0 +1,26 @@
...
+        public static InstalledModuleInfo ToWebModel(this coreModel.ManifestModuleInfo moduleInfo)
+        {
+            var retVal = new InstalledModuleInfo();
+            retVal.InjectFrom(moduleInfo);
+            retVal.Version = moduleInfo.Version.ToString();
+            retVal.PlatformVersion = moduleInfo.PlatformVersion.ToString();
+            retVal.Groups = moduleInfo.Groups;
+            if (moduleInfo.Dependencies != null)
+            {
+                retVal.Dependencies = moduleInfo.Dependencies.Select(x => new ModuleIdentity { Id = x.Id, Version = x.Version.ToString() }).ToList();
```
t13ka: I guess we shouldn't use not readable variables in queries (like x, a, b.. etc). A much better use name of the variable in the singular, eg: .Dependencies.Select(**dependency** => new...)

VirtoCommerce.Platform.Web/Scripts/app/workspace.tpl.html
```diff
@@ -1,10 +1,11 @@
-﻿<va-main-menu ng-model="mainMenu" on-menu-changed="onMainMenuChanged(mainMenu)">
+<va-main-menu ng-model="mainMenu" on-menu-changed="onMainMenuChanged(mainMenu)">
     <div class="info">
         <a class="info-license" ng-if="!license" href="" ng-click="showLicense()">{{'platform.templates.licensing.license-absent' | translate}}</a>
         <a class="info-license" ng-if="license" href="" ng-click="showLicense()" ng-class="{'__valid': !license.hasExpired}">{{'platform.templates.licensing.license-present' | translate: license}}</a>
     </div>
     <div class="info">
         <span class="info-version">{{platformVersion}}</span>
+        <a class="info-license" href="" ng-click="showVersionInfo()">Platform Info</a>
```
t13ka: Why the text "Platform Info" is not localized?

VirtoCommerce.Platform.Web/Scripts/app/systemInfo/blades/system-info.js
```diff
@@ -0,0 +1,56 @@
...
+            var blade = $scope.blade;
+
+            function stringify(data) {
+                return JSON.stringify(data, null, "\t");
+            }
+
+            blade.initializeBlade = function () {
+                diagnostics.getSystemInfo({}, function (results) {
+                    blade.isLoading = false;
+                    blade.title = "Platform Info";
```
t13ka: Why the text "Platform Info" is not localized here?

VirtoCommerce.Platform.Web/Converters/Diagnostics/InstalledModuleInfoConverter.cs
yecli: We have `x` as default for lambda expressions in our coding style now. For one variable - it is `x`. For more - we are using meaningful names. It is the current practice.
Though it is not reflected in our style guide, created the issue on it: https://github.com/VirtoCommerce/styleguide/issues/4

t13ka: Yes, you right in that the code will be shorter, but
It is not good practice for me, because:
1. Very often we have long query expressions and quite difficult to catch the real meaning of the variable if this var have name 'x' and etc.
2. If we have an iteration(like in the mentioned code) then each iteration item should have the name that is showing the singular meaning. 
3. If we doing the math calculation in the same code then we might get the name collisions.
4. The rules that you mentioned might be not obvious for common developers. It's +1 rule for memorizing.

yecli: There could be pros and cons for any variant. We have this style accepted now. Maybe it could be changed after the team discussion, but I think we should follow our conventions before the change.

t13ka: Then we shouldn't add only Artem into reviewers for this, isn't? :)
https://github.com/VirtoCommerce/styleguide/issues/4

yecli: Sure, you are free to add any VC members to the reviewers.
Code style discussion took place before, and with the issue, I'm asking Atrem to add this particular item in the way I remember it to the style guide.
To introduce another rule (or write it for the first time "to paper") I think it is better to raise the question and discuss it in e.g. our backend developers channel when you can ask whole VC dev team about this.

VirtoCommerce.Platform.Web/Model/Diagnostics/InstalledModuleInfo.cs
```diff
@@ -0,0 +1,24 @@
+using System.Collections.Generic;
+using VirtoCommerce.Platform.Web.Model.Modularity;
+
+namespace VirtoCommerce.Platform.Web.Model.Diagnostics
+{
+    public class InstalledModuleInfo
```
tatarincev: Why we need this new class? Why we can't use exists [ModuleDescriptor](https://github.com/VirtoCommerce/vc-platform/blob/master/VirtoCommerce.Platform.Web/Model/Modularity/ModuleDescriptor.cs)? Especially it has the same signature

VirtoCommerce.Platform.Web/Converters/Diagnostics/InstalledModuleInfoConverter.cs
```diff
@@ -0,0 +1,26 @@
+using System.Linq;
+using Omu.ValueInjecter;
+using VirtoCommerce.Platform.Web.Model.Diagnostics;
+using VirtoCommerce.Platform.Web.Model.Modularity;
+using coreModel = VirtoCommerce.Platform.Core.Modularity;
+
+namespace VirtoCommerce.Platform.Web.Converters.Diagnostics
+{
+    public static class InstalledModuleInfoConverter
```
tatarincev: If we reused this exits [ModuleDescriptor](https://github.com/VirtoCommerce/vc-platform/blob/master/VirtoCommerce.Platform.Web/Model/Modularity/ModuleDescriptor.cs) we wouldn't need to define the new converter and would use this one [ModuleDescriptorConverter](https://github.com/VirtoCommerce/vc-platform/blob/master/VirtoCommerce.Platform.Web/Converters/Modularity/ModuleDescriptorConverter.cs)

VirtoCommerce.Platform.Web/Model/Diagnostics/PlatformInfo.cs
```diff
@@ -0,0 +1,7 @@
+namespace VirtoCommerce.Platform.Web.Model.Diagnostics
+{
+    public class PlatformInfo
```
tatarincev: I think this class is redundant, you can use simple string for platform version representation

VirtoCommerce.Platform.Web/Converters/Diagnostics/InstalledModuleInfoConverter.cs
yecli: Added PR about the current convention to have the starting point: https://github.com/VirtoCommerce/styleguide/pull/5.

VirtoCommerce.Platform.Web/Model/Diagnostics/InstalledModuleInfo.cs
trueboroda: Yes. I agree.

VirtoCommerce.Platform.Web/Converters/Diagnostics/InstalledModuleInfoConverter.cs
trueboroda: I agree.

VirtoCommerce.Platform.Web/Scripts/app/workspace.tpl.html
trueboroda: This is fixed

VirtoCommerce.Platform.Web/Scripts/app/systemInfo/blades/system-info.js
trueboroda: This is fixed

VirtoCommerce.Platform.Web/Licensing/LicenseProvider.cs
```diff
@@ -0,0 +1,24 @@
...
+                license = License.Parse(rawLicense);
+
+                if (license != null)
+                {
+                    license.RawLicense = null;
+                }
+            }
+            return license;
+        }
+    }
```
t13ka: Not testable code. Violation of the SOLID principles. What if we want to get the license from a different source?
I'm really hoping that we'll fix violations like this in future )

VirtoCommerce.Platform.Web/Controllers/HomeController.cs
```diff
@@ -27,7 +26,7 @@ public ActionResult Index()
             var version = PlatformVersion.CurrentVersion.ToString();
             var demoCredentials = ConfigurationHelper.GetAppSettingsValue("VirtoCommerce:DemoCredentials");
             var resetTimeStr = ConfigurationHelper.GetAppSettingsValue("VirtoCommerce:DemoResetTime");
-            var license = LoadLicense();
+            var license = LicenseProvider.LoadLicense();
```
t13ka: If we have a "License Provider" then would be enough to name this method just "Load"

akak1977:
Released: https://github.com/VirtoCommerce/vc-platform/releases/tag/v2.13.54

PR #1738, opened 2019-12-04
#vp 961 export to pdf on linux

kostyrin:


src/VirtoCommerce.Platform.Core/PlatformOptions.cs
```diff
@@ -26,6 +25,7 @@ public class PlatformOptions
         //Default path to store export files 
         public string DefaultExportFolder = "App_Data/Export/";
         public string DefaultExportFileName = "exported_data.zip";
+        public string LibraryPath { get; set; }
```
akak1977: I think there is need default value. Otherwise it will fail in case of somebody forget to add line to appsettings.json

src/VirtoCommerce.Platform.Web/Extensions/ServiceCollectionExtensions.cs
```diff
@@ -107,5 +110,30 @@ public static IServiceCollection AddSecurityServices(this IServiceCollection ser
 
             return services;
         }
+
+
+
+        public static IServiceCollection AddLibraries(this IServiceCollection services)
```
akak1977: It seems will be good to rename to AddUnmanagedLibraries or AddNativeLibraries .

src/VirtoCommerce.Platform.Modules/AssemblyLoading/CustomAssemblyLoadContext.cs
```diff
@@ -0,0 +1,22 @@
+using System;
+using System.Reflection;
+using System.Runtime.Loader;
+
+namespace VirtoCommerce.Platform.Modules.AssemblyLoading
+{
+    public class CustomAssemblyLoadContext : AssemblyLoadContext
```
tatarincev: Why do we need this type? It doesn't implement anything.

src/VirtoCommerce.Platform.Web/Extensions/ServiceCollectionExtensions.cs
tatarincev: I would like to recommend move this logic into the new class `UnmanagedLibraryLoader` and encapsulate all this logic there.

src/VirtoCommerce.Platform.Web/Extensions/ServiceCollectionExtensions.cs
```diff
@@ -107,5 +110,30 @@ public static IServiceCollection AddSecurityServices(this IServiceCollection ser
...
+        {
+            var providerSnapshot = services.BuildServiceProvider();
+            var assemblyResolver = providerSnapshot.GetRequiredService<IAssemblyResolver>();
+            var hostingEnvironment = providerSnapshot.GetRequiredService<IWebHostEnvironment>();
+            var platformOptions = providerSnapshot.GetService<IOptions<PlatformOptions>>();
+
+            var librariesDir = Path.Combine(hostingEnvironment.ContentRootPath, platformOptions.Value.LibraryPath);
```
tatarincev: Library path can be resolved in startup https://github.com/VirtoCommerce/vc-platform/blob/release/3.0.0/src/VirtoCommerce.Platform.Web/Startup.cs#L370

src/VirtoCommerce.Platform.Modules/AssemblyLoading/CustomAssemblyLoadContext.cs
kostyrin: because there is protected method LoadUnmanagedDllFromPath in the base class AssemblyLoadContext, look at (link)[https://docs.microsoft.com/en-us/dotnet/standard/assembly/unloadability]

src/VirtoCommerce.Platform.Core/ProcessSettings/WkHtmlToPdfSettings.cs
```diff
@@ -0,0 +1,13 @@
+using System.Collections.Generic;
+
+namespace VirtoCommerce.Platform.Core.ProcessSettings
+{
+    public class WkHtmlToPdfSettings : ProcessSettings
```
tatarincev: I think this setting must be moved from there into modules where it is used.  For test purposes, you can define typed settings for a required command directly in the test solution.

PR #1754, opened 2019-12-20
VP-1161 Creating dynamic properties in Store

kostyrin:
https://virtocommerce.atlassian.net/browse/VP-1161

### DIFF
--- a/VirtoCommerce.Platform.Web/Scripts/app/dynamicProperties/blades/propertyValue-list.js
+++ b/VirtoCommerce.Platform.Web/Scripts/app/dynamicProperties/blades/propertyValue-list.js
@@ -67,7 +67,6 @@
         if (isDirty()) {
             angular.copy(blade.currentEntities, blade.data.dynamicProperties);
             angular.copy(blade.currentEntities, blade.origEntity);
-            blade.currentEntity.dynamicProperties = blade.data.dynamicProperties;
         }
         $scope.bladeClose();
     };
--- a/VirtoCommerce.Platform.Web/Scripts/app/dynamicProperties/blades/propertyValue-list.js
+++ b/VirtoCommerce.Platform.Web/Scripts/app/dynamicProperties/blades/propertyValue-list.js
@@ -42,10 +42,10 @@
     };
 
     function refreshDynamicProperties(data)  { 
-        _.each(data, function (x) {
-            var existProperty = _.find(blade.currentEntity.dynamicProperties, function (y) { return y.id == x.id; });
+        _.each(data, function (dataItem) {
+            var existProperty = _.find(blade.currentEntity.dynamicProperties, function (property) { return property.id == dataItem.id; });
             if (!angular.isDefined(existProperty)) {
-                blade.currentEntity.dynamicProperties.push(x);
+                blade.currentEntity.dynamicProperties.push(dataItem);
             }
         });
     }
--- a/VirtoCommerce.Platform.Web/Scripts/app/dynamicProperties/blades/propertyValue-list.js
+++ b/VirtoCommerce.Platform.Web/Scripts/app/dynamicProperties/blades/propertyValue-list.js
@@ -9,46 +9,48 @@
 
     blade.refresh = function () {
         blade.data = blade.currentEntity;
-        var rawProperties = angular.copy(blade.currentEntity.dynamicProperties);
+        dynamicPropertiesApi.getAll({typeName: blade.currentEntity.objectType}, function (response) {
+            var rawProperties = response;
 
-        _.each(rawProperties, function (x) {
-            if (x.values) {
-                x.values.sort(function (a, b) {
-                    return a.value && b.value
-                        ? (a.value.name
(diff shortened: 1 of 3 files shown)


VirtoCommerce.Platform.Web/Scripts/app/dynamicProperties/blades/propertyValue-list.js
```diff
@@ -12,15 +12,17 @@
...
-                return a.value && b.value
-                    ? (a.value.name
-                        ? a.value.name.localeCompare(b.value.name)
-                        : angular.isString(a.value) && angular.isString(b.value)
-                            ? a.value.localeCompare(b.value)
-                            : a.value < b.value ? -1 : a.value > b.value ? 1 : 0)
-                    : -1;
-            });
+            if (x.values) {
+                x.values.sort(function (a, b) {
```
t13ka: not readable code. It is quite difficult to read and understand code like this. I think we should refactor similar code, especially if we are trying to solve some issues which are close to the context.
Why we have here the "magic numbers"?
Why we don't simplify and split some operators here for better readability?
Example:
if (x.values) {
                x.values.sort(function (a, b) {

                  var minusOne = -1; // what is it?
                  var one = 1; // what is it?
                  var zero = 0; // again..
                  var valueA = a.value;
                  var valueB = b.value;
                  var isValueAString = angular.isString(valueA); 
                  var isValueBString = angular.isString(valueB);
                  var nameA = valueA.name;
                  var nameB = valueB.name;
                  var isAEqualsB = nameA.localeCompare(nameB);
                  var isNameAEqualsValueB = valueA.localeCompare(valueB);

                  // here the main logic
                  var someVariable = (nameA ? isAEqualsB : isValueAString && isValueBString
                         ? isNameAEqualsValueB
                         : valueA < valueB ? minusOne : valueA > valueB ? one : zero);
                         
                    return valueA && valueB ? someVariable : minusOne;
                });
            }

It's just an example. I'm sure that here we can apply further decomposition )

VirtoCommerce.Platform.Web/Scripts/app/dynamicProperties/blades/propertyValue-list.js
```diff
@@ -39,6 +41,15 @@
         blade.isLoading = false;
     };
 
+    function refreshDynamicProperties(data)  { 
+        _.each(data, function (x) {
+            var existProperty = _.find(blade.currentEntity.dynamicProperties, function (y) { return y.id == x.id; });
+            if (!angular.isDefined(existProperty)) {
+                blade.currentEntity.dynamicProperties.push(x);
+            }
```
t13ka: Why we can't use here more clear names of variables in functions?
Instead 'x' => might be 'dataItem'
Instead 'y' => can be 'property'
etc

VirtoCommerce.Platform.Web/Scripts/app/dynamicProperties/blades/propertyValue-list.js
kostyrin: I agree need to refactor but I added just a condition "if" for safety.. inside it is legacy code. line (15-25)

t13ka: Ok. So be it ) We'll talk about it later

VirtoCommerce.Platform.Web/Scripts/app/dynamicProperties/blades/propertyValue-list.js
```diff
@@ -39,6 +41,15 @@
         blade.isLoading = false;
     };
 
+    function refreshDynamicProperties(data)  {
```
tatarincev: This implementation looks like as `ad hoc`  solution, better to implement the propagation of the changes up to blades tree as it has done in the third version and in addition, needs to change dynamic properties binding, need to get dynamic properties meta information from API rather than  from context object

kostyrin: made some chenges

VirtoCommerce.Platform.Web/Scripts/app/dynamicProperties/blades/propertyValue-list.js
```diff
@@ -9,47 +9,49 @@
...
+
+            _.each(rawProperties, function(property) {
+                property.values = [];
+                var filteredProperty = _.find(blade.currentEntity.dynamicProperties, function (item) { return item.id === property.id; });
+                if (filteredProperty) {
+                    property.values = filteredProperty.values;
+                }
+            })
+
+            _.each(rawProperties, function (x) {
```
t13ka: 'x' - is bad for use, 'rawProperty' - better

kostyrin: I agree! this is also legacy code

PR #1764, opened 2020-01-10
dev 1634 Localization of custom modules is not working

AliveMen:
Add server localization codes to web.config
Copy satellite localization assemblies from modules

### DIFF
--- a/VirtoCommerce.Platform.Web/Modularity/ManifestModuleCatalog.cs
+++ b/VirtoCommerce.Platform.Web/Modularity/ManifestModuleCatalog.cs
@@ -42,9 +42,9 @@ namespace VirtoCommerce.Platform.Web.Modularity
             {
                 Directory.CreateDirectory(_assembliesPath);
             }
-
-            if (!contentPhysicalPath.EndsWith("\\", StringComparison.OrdinalIgnoreCase))
-                contentPhysicalPath += "\\";
+            var separator = Path.DirectorySeparatorChar.ToString();
+            if (!contentPhysicalPath.EndsWith(separator, StringComparison.OrdinalIgnoreCase))
+                contentPhysicalPath += separator;
 
             var rootUri = new Uri(contentPhysicalPath);
 
@@ -180,8 +180,9 @@ namespace VirtoCommerce.Platform.Web.Modularity
                 {
                     if (!string.IsNullOrEmpty(languageCode))
                     {
-                        var localizationDirectoryPath = Path.Combine(sourceDirectoryPath, $"{languageCode}\\");
-                        var localizationTargetDirectoryPath = Path.Combine(targetDirectoryPath, $"{languageCode}\\");
+                        var separator = Path.DirectorySeparatorChar;
+                        var localizationDirectoryPath = Path.Combine(sourceDirectoryPath, $"{languageCode}{separator}");
+                        var localizationTargetDirectoryPath = Path.Combine(targetDirectoryPath, $"{languageCode}{separator}");
                         if (Directory.Exists(localizationDirectoryPath))
                         {
                             var localizationDirectoryUri = new Uri(localizationDirectoryPath);
@@ -212,7 +213,8 @@ namespace VirtoCommerce.Platform.Web.Modularity
         {
             if (sourceParentPath != null)
             {
-                var sourceDirectoryPath = Path.Combine(sourceParentPath, "bin\\");
+                var separator = Path.DirectorySeparatorChar;
+                var sourceDirectoryPath = Path.Combine(sourceParentPath, $"bin{separator}");
 
                 CopyLocalizationAssemblies(sourceDirectoryPath, targetDirectoryPath);
 
--- a/VirtoCommerce.Platform.Web/Modularity/ManifestModuleCatalog.cs
+++ b/VirtoCommerce.Platform.Web/Modularity/ManifestModuleCatalog.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using System.Collections.Generic;
 using System.Diagnostics;
 using System.IO;
@@ -170,11 +170,51 @@ namespace VirtoCommerce.Platform.Web.Modularity
             return result;
         }
 
+        private static void CopyAssembliesLocalizations(string sourceDirectoryPath, string targetDirectoryPath)
+        {
+
+            var languageCodes = ConfigurationHelper.GetAppSettingsValue("VirtoCommerce:Localization:ServerLocalizationCodes");
+            foreach (var languageCode in languageCodes.Split(','))
+            {
+                try
+                {
+                    if (!string.IsNullOrEmpty(languageCode))
+                    {
+                        var localizationDirectoryPath = Path.Combine(sourceDirectoryPath, $"{languageCode}\\");
+                        var localizationTargetDirectoryPath = Path.Combine(targetDirectoryPath, $"{languageCode}\\");
+                        if (Directory.Exists(localizationDirectoryPath))
+                        {
+                            var localizationDirectoryUri = new Uri(localizationDirectoryPath);
+
+                            foreach (var localizationFilePath in Directory.EnumerateFiles(localizationDirectoryPath))
+                            {
+                                if (IsAssemblyFile(localizationFilePath))
+                                {
+                                    var relativePath = MakeRelativePath(localizationDirectoryUri, localizationFilePath);
+                                    var targetFilePath = Path.Combine(localizationTargetDirectoryPath, relativePath);
+
+                                    CopyFile(localizationFilePath, targetFilePath);
+                                }
+                            }
+                        }
+                    }
+                }
+
+                catch
+                {
+                    // IDLE
+                }
+            }
+
+        }
+
         private static void CopyAssemblies(string sourceParentPath, string targetDirectoryPath)
         {
             if (sourceParentPath != null)
             {
                 var sourceDirectoryPath = Path.Combine(sourceParentPath, "bin\\");
+
+                CopyAssembliesLocalizations(sourceDirectoryPath, targetDirectoryPath);
 
                 if (Directory.Exists(sourceDirectoryPath))
                 {
--- a/VirtoCommerce.Platform.Web/Modularity/ManifestModuleCatalog.cs
+++ b/VirtoCommerce.Platform.Web/Modularity/ManifestModuleCatalog.cs
@@ -172,8 +172,8 @@ namespace VirtoCommerce.Platform.Web.Modularity
 
         private static void CopyLocalizationAssemblies(string sourceDirectoryPath, string targetDirectoryPath)
         {
-
             var languageCodes = ConfigurationHelper.GetAppSettingsValue("VirtoCommerce:Localization:ServerLocalizationCodes");
+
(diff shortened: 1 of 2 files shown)


VirtoCommerce.Platform.Web/Modularity/ManifestModuleCatalog.cs
```diff
@@ -170,25 +170,50 @@ public override void Validate()
...
-                {
-                    var sourceDirectoryUri = new Uri(sourceDirectoryPath);
+                CopyLocalizationAssemblies(sourceDirectoryPath, targetDirectoryPath);
+
+                CopyAssemblyFiles(targetDirectoryPath, sourceDirectoryPath);
+            }
+        }
 
-                    foreach (var sourceFilePath in Directory.EnumerateFiles(sourceDirectoryPath))
+        private static void CopyAssemblyFiles(string targetDirectoryPath, string sourceDirectoryPath)
```
yecli: Lets order method parameters in the following way (as in other places):
source, target

VirtoCommerce.Platform.Web/Modularity/ManifestModuleCatalog.cs
```diff
@@ -170,25 +170,50 @@ public override void Validate()
             return result;
         }
 
+        private static void CopyLocalizationAssemblies(string sourceDirectoryPath, string targetDirectoryPath)
+        {
+            var languageCodes = ConfigurationHelper.GetAppSettingsValue("VirtoCommerce:Localization:ServerLocalizationCodes");
```
yecli: Need to provide default value "string.Empty" here, to avoid NullReference.
Maybe there is a method in `ConfigurationHelper` that gets value or default.

VirtoCommerce.Platform.Web/Web.config
```diff
@@ -57,7 +57,7 @@
         <add key="VirtoCommerce:EnableBundlesOptimizations" value="false"/>
         <add key="VirtoCommerce:AutoInstallModuleBundles" value="commerce"/>
         <!-- Use this option to allow CORS from these origins. Examples: "http://domain1.com;https://domain2.org" -->
-        <add key="VirtoCommerce:CORS:AllowedOrigins" value=""/>
+        <add key="VirtoCommerce:CORS:AllowedOrigins" value="*"/>
```
yecli: Revert it

VirtoCommerce.Platform.Web/Web.config
```diff
@@ -122,6 +122,8 @@
         <!--<add key="GoogleTagManager:ContainerId" value="SECRET" />-->
         <!--Use this setting to specify redis channel name to be used by cache. Only instances with the same channel name will recieve messages. -->
         <!--<add key="VirtoCommerce:Cache:Redis:ChannelName" value="(Place name of desired channel here)" />-->
+        <!-- Copy satellite localization assemblies from module bin folder with language code names: Example "de-DE,ru-RU"-->
```
yecli:
```
<!--Use this setting to support server localization resources. Satellite localization assemblies for the specified languages-->
<!-- from the given comma-separated list will be copied to the probing folder.-->
<!--Example value: "de-DE,ru-RU".-->
```

VirtoCommerce.Platform.Web/Web.config
```diff
@@ -122,6 +122,8 @@
...
+        <add key="VirtoCommerce:Localization:ServerLocalizationCodes" value=""/>
```
yecli: Let's name this setting `VirtoCommerce:Resources:CultureNames`

VirtoCommerce.Platform.Web/Modularity/ManifestModuleCatalog.cs
yecli: Also please name it `cultureNames`

VirtoCommerce.Platform.Web/Modularity/ManifestModuleCatalog.cs
AliveMen: fixed

VirtoCommerce.Platform.Web/Modularity/ManifestModuleCatalog.cs
AliveMen: done

VirtoCommerce.Platform.Web/Web.config
AliveMen: done

VirtoCommerce.Platform.Web/Web.config
AliveMen: done

VirtoCommerce.Platform.Web/Web.config
AliveMen: done

VirtoCommerce.Platform.Web/Modularity/ManifestModuleCatalog.cs
```diff
@@ -170,25 +170,50 @@ public override void Validate()
             return result;
         }
 
-        private static void CopyAssemblies(string sourceParentPath, string targetDirectoryPath)
+        private static void CopyLocalizationAssemblies(string sourceDirectoryPath, string targetDirectoryPath)
+        {
+            var cultureNames = ConfigurationHelper.GetAppSettingsValue("VirtoCommerce:Resources:CultureNames") ?? string.Empty;
+
```
yecli: Please make a lazy property for this option to read it only once.

VirtoCommerce.Platform.Web/Modularity/ManifestModuleCatalog.cs
```diff
@@ -42,9 +42,9 @@ protected override void InnerLoad()
             {
                 Directory.CreateDirectory(_assembliesPath);
             }
-
-            if (!contentPhysicalPath.EndsWith("\\", StringComparison.OrdinalIgnoreCase))
-                contentPhysicalPath += "\\";
+            var separator = Path.DirectorySeparatorChar.ToString();
+            if (!contentPhysicalPath.EndsWith(separator, StringComparison.OrdinalIgnoreCase))
+                contentPhysicalPath += separator;
```
t13ka: braces are missed

VirtoCommerce.Platform.Web/Modularity/ManifestModuleCatalog.cs
```diff
@@ -170,25 +170,50 @@ public override void Validate()
...
+        {
+            var cultureNames = ConfigurationHelper.SplitAppSettingsStringValue("VirtoCommerce:Resources:CultureNames");
+
+            foreach (var languageCode in cultureNames)
+            {
+                if (!string.IsNullOrEmpty(languageCode))
+                {
+                    var separator = Path.DirectorySeparatorChar;
+                    var localizationDirectoryPath = Path.Combine(sourceDirectoryPath, $"{languageCode}{separator}");
+                    var localizationTargetDirectoryPath = Path.Combine(targetDirectoryPath, $"{languageCode}{separator}");
```
t13ka: maybe names "directoryPath" and "targetDirectoryPath" be enough for these variables?

VirtoCommerce.Platform.Web/Modularity/ManifestModuleCatalog.cs
```diff
@@ -170,25 +170,50 @@ public override void Validate()
             return result;
         }
 
+        private static void CopyLocalizationAssemblies(string sourceDirectoryPath, string targetDirectoryPath)
```
t13ka: Can we name this just "CopyAssamblies"?

VirtoCommerce.Platform.Web/Modularity/ManifestModuleCatalog.cs
```diff
@@ -170,25 +170,50 @@ public override void Validate()
...
+                CopyLocalizationAssemblies(sourceDirectoryPath, targetDirectoryPath);
+
+                CopyAssemblyFiles(sourceDirectoryPath, targetDirectoryPath);
+            }
+        }
 
-                    foreach (var sourceFilePath in Directory.EnumerateFiles(sourceDirectoryPath))
+        private static void CopyAssemblyFiles(string sourceDirectoryPath, string targetDirectoryPath)
+        {
+            if (Directory.Exists(sourceDirectoryPath))
```
t13ka: try to invert this, probably your code will be shorter without extra braces

VirtoCommerce.Platform.Web/Modularity/ManifestModuleCatalog.cs
AliveMen: This app setting removed

VirtoCommerce.Platform.Web/Modularity/ManifestModuleCatalog.cs
AliveMen: done

VirtoCommerce.Platform.Web/Modularity/ManifestModuleCatalog.cs
AliveMen: The algorithm had been changed. Names had been simplified a bit.

VirtoCommerce.Platform.Web/Modularity/ManifestModuleCatalog.cs
AliveMen: I don't think so. This method copy only resource localisation assemblies. Method CopyAssemblyFiles copy all other assemblies.

VirtoCommerce.Platform.Web/Modularity/ManifestModuleCatalog.cs
AliveMen: I agree with you. But we use a single return code convention here.

VirtoCommerce.Platform.Web/Modularity/ManifestModuleCatalog.cs
```diff
@@ -170,25 +173,59 @@ public override void Validate()
...
+        private static void CopyLocalizationAssemblies(string sourceDirectoryPath, string targetDirectoryPath)
+        {
+            if (Directory.Exists(sourceDirectoryPath))
+            {
+                foreach (var sourceFilePath in Directory.EnumerateFiles(sourceDirectoryPath, "*.resources.dll", SearchOption.AllDirectories))
+                {
+                    var separator = Path.DirectorySeparatorChar;
+                    var directory = Path.GetDirectoryName(sourceFilePath);
+
+                    var languageCode = directory?.Split(separator).Last();
```
yecli: Here a bug could be introduced. 
I think we need to ensure we have only one directory in source resource DLL path, otherwise path localization file from path like "/A/B/en-US/" will be copied (probably should not). And then fullSourceDirectoryPath will be wrong.
You could use first level subdirectory search
```var files = 
    Directory.EnumerateDirectories(root).SelectMany(
        directory => Directory.EnumerateFiles(directory, "*.resources.dll"));
```
Seems `languageCode` and manual path construction will not be needed at all.

AliveMen: done

yecli:
Versions conflict resolution works as expected - as for ordinary DLL's.

PR #1782, opened 2020-02-07
VP-1312 Migrate AvalaraTax to v3

AliveMen:
Add options for Avalara credentials


yecli:
No need to add to default config, please add example to PR

PR #1810, opened 2020-03-23
SLint-related codestyle changes

akak1977:


src/VirtoCommerce.Platform.Core/PlatformConstants.cs
```diff
@@ -67,7 +67,7 @@ public static class Settings
         {
             public static class Security
             {
-                public static SettingDescriptor SecurityAccountTypes = new SettingDescriptor
+                public static SettingDescriptor SecurityAccountTypes { get; set; } = new SettingDescriptor
```
tatarincev: left only `get;`  accessors for constants.

src/VirtoCommerce.Platform.Core/Common/ReflectionUtility.cs
```diff
@@ -86,7 +85,7 @@ public static bool IsDerivativeOf(this Type type, Type typeToCompare)
         {
             if (type == null)
             {
-                throw new NullReferenceException();
+                throw new ArgumentNullException("type");
```
tatarincev: use `nameof()` instead of string

akak1977: OK

src/VirtoCommerce.Platform.Core/PlatformConstants.cs
akak1977: OK

PR #1822, opened 2020-04-06
VP-1838 sonar bug remove base

AliveMen:
Fix sonar issue

### DIFF
--- a/src/VirtoCommerce.Platform.Core/Domain/Entity.cs
+++ b/src/VirtoCommerce.Platform.Core/Domain/Entity.cs
@@ -33,7 +33,7 @@ namespace VirtoCommerce.Platform.Core.Common
         {
             unchecked
             {
-                return IsTransient() ? base.GetHashCode() : Id.GetHashCode();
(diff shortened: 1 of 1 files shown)


src/VirtoCommerce.Platform.Core/Domain/Entity.cs
```diff
@@ -33,7 +33,7 @@ public override int GetHashCode()
         {
             unchecked
             {
-                return IsTransient() ? base.GetHashCode() : Id.GetHashCode();
+                return IsTransient() ? GetHashCode() : Id.GetHashCode();
```
yecli, line 36: It will result in stack overflow, as the method will call himself in case of the Transient object.
The previous implementation was ok. Need to suppress the warning.

AliveMen:
The previous implementation was ok

src/VirtoCommerce.Platform.Core/Domain/Entity.cs
AliveMen: Warning suppressed.

PR #1823, opened 2020-04-07
Feature/3.0.0/https config

Woland2k:
### Problem
Platform can't run on both http and https ports
Platform fails to create correct folders on linux environment

### Solution
Add config parameter to enable running on http
Fix to filestorageprovider to create correct directories under linux


Woland2k:
Can we get this merged?

PR #1830, opened 2020-04-13
VP-1921: Add environment info to API

yecli:
- Added info about the environment to the system info;
- Fixed return type in '/api/platform/diagnostics/systemInfo';

### DIFF
--- a/src/VirtoCommerce.Platform.Web/Controllers/Api/DiagnosticsController.cs
+++ b/src/VirtoCommerce.Platform.Web/Controllers/Api/DiagnosticsController.cs
@@ -4,7 +4,6 @@ using System;
 using System.Linq;
 using Microsoft.AspNetCore.Authorization;
 using Microsoft.AspNetCore.Mvc;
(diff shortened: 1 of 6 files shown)


src/VirtoCommerce.Platform.Web/Controllers/Api/DiagnosticsController.cs
```diff
@@ -43,5 +46,20 @@ public ActionResult<License> GetSystemInfo()
             };
             return Ok(result);
         }
+
+        [HttpGet]
+        [Route("environmentInfo")]
+        [Authorize(PlatformConstants.Security.Permissions.EnvironmentInfoQuery)]
+        public ActionResult<EnvironmentInfo> GetEnvironmentInfo()
```
tatarincev, line 53: ENV information must be part of the SystemInfo type and retrieved using the GetSystemInfo method.

yecli: Done

PR #1831, opened 2020-04-14
Added retry interval setting for failed notifications.

tpitlik:
### Problem
At the moment retry interval for failed notifications is 5 hours by default and there is no possibility to change it.

### Solution
Add ability to set in admin panel retry interval for resending failed notifications.

### Proposed of changes
Add new parameter for retry interval that can be changed in module settings.

### DIFF
--- a/VirtoCommerce.Platform.Core/Notifications/SearchNotificationCriteria.cs
+++ b/VirtoCommerce.Platform.Core/Notifications/SearchNotificationCriteria.cs
@@ -6,6 +6,10 @@ namespace VirtoCommerce.Platform.Core.Notifications
 {
 	public class SearchNotificationCriteria : ValueObject
     {
+        public SearchNotificationCriteria()
+        {
+            RepeatMinutesIntervalForFail = 5 * 60;
+        }
         public int Take { get; set; }
 		public int Skip { get; set; }
         /// <summary>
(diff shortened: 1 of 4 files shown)


VirtoCommerce.Platform.Core/Notifications/SearchNotificationCriteria.cs
```diff
@@ -5,11 +5,7 @@ namespace VirtoCommerce.Platform.Core.Notifications
 {
 	public class SearchNotificationCriteria : ValueObject
     {
-        public SearchNotificationCriteria()
```
tatarincev, line 8: To avoid breaking changes effect introduced by these changes I would like to recommend preserving previous default value for RepeatMinutesIntervalForFail property 5*60, and do not remove `RepeatHoursIntervalForFail ` and mark it as obsolete

VirtoCommerce.Platform.Core/Notifications/SearchNotificationCriteria.cs
```diff
@@ -1,15 +1,12 @@
...
 namespace VirtoCommerce.Platform.Core.Notifications
 {
 	public class SearchNotificationCriteria : ValueObject
     {
-        public SearchNotificationCriteria()
-        {
-            RepeatHoursIntervalForFail = 5;
-        }
-		public int Take { get; set; }
+        public int Take { get; set; }
```
tatarincev, line 9: you forgot to add  the default value for  RepeatMinutesIntervalForFail property (previous value was 5 hours)

yecli:
Released: https://github.com/VirtoCommerce/vc-platform/releases/tag/v2.13.62

PR #1833, opened 2020-04-15
Fix IE-11 compatibility

yecli:


yecli:
Released: https://github.com/VirtoCommerce/vc-platform/releases/tag/v2.13.62

PR #1843, opened 2020-04-22
VP-1883: Build webpack if output files do not exist

AliveMen:
VP-1883: Add js script for initial webpack build and bind it to the build project action


Woland2k:
WebPack doesn't build when project is first opened, possibly due to this error? https://github.com/VirtoCommerce/vc-platform/issues/1854 @AliveMen  @yecli

yecli:
Confirming that it works for VS 16.5.1 and not working for 16.5.4 with this error:
image

Let's wait for the VS update and check again then.

PR #1865, opened 2020-05-14
VP-2050: Enrich TypeExtensions with deep clone independence check methods

akak1977:


src/VirtoCommerce.Platform.Core/Extensions/TypeExtensions.cs
```diff
@@ -67,5 +71,86 @@ private static string PrettyPrintRecursive(Type type, int depth)
...
+        {
+            if (original != null && expected != null)
+            {
+                var typeOfOriginal = original.GetType();
+                if (!IsPrimitive(typeOfOriginal) && !visited.Contains(original))
+                {
+                    visited.Add(original);
+                    if (ReferenceEquals(original, expected))
+                    {
+                        throw new MemberAccessException(@$"Deep clone check failed: objects at path {memberPath} are reference equal.");
```
yecli, line 117: Maybe we could throw one type of exception to simplify assert using.

src/VirtoCommerce.Platform.Core/Extensions/TypeExtensions.cs
```diff
@@ -67,5 +71,86 @@ private static string PrettyPrintRecursive(Type type, int depth)
...
+        /// Ensures no shared references with expected object (original and expected objects are fully independent from each other).
+        /// </summary>
+        /// <param name="original"></param>
+        /// <param name="expected"></param>
+        public static void AssertNoSharedRefsWith(this object original, object expected)
+        {
+            AssertNoSharedRefsWith(original, expected, new List<object>(), original.GetType().Name);
+        }
+
+        private static void AssertNoSharedRefsWith(object original, object expected, List<object> visited, string memberPath)
```
tatarincev, line 107: maybe better would be narrow down this method responsibility?
Rename this method to something like this  IEnumerable<object> GetAllObjectPropertiesRecurively(obj) where return all  found properties values via `yield return  propValue` expression.

src/VirtoCommerce.Platform.Core/Extensions/TypeExtensions.cs
```diff
@@ -67,5 +68,74 @@ private static string PrettyPrintRecursive(Type type, int depth)
                 : $"{nameParts[0]}<{string.Join(",", genericArguments.Select(t => PrettyPrintRecursive(t, depth + 1)))}>";
         }
 
+
+        /// <summary>
+        /// Iterates recursively over each public property of object to gather member values.
+        /// </summary>
+        /// <param name="original"></param>
+        /// <returns></returns>
+        public static IEnumerable<KeyValuePair<string, object>> ObjectWalker(this object original)
```
tatarincev, line 77: I would like to suggest renaming to `TraverseObjectGraph` because the method name must contain a verb.

src/VirtoCommerce.Platform.Core/Extensions/TypeExtensions.cs
```diff
@@ -67,5 +68,74 @@ private static string PrettyPrintRecursive(Type type, int depth)
...
+        {
+            foreach (var result in original.ObjectWalker(new List<object>(), original.GetType().Name))
+            {
+                yield return result;
+            }
+        }
+
+        private static IEnumerable<KeyValuePair<string, object>> ObjectWalker(this object obj, List<object> visited, string memberPath)
```
tatarincev, line 85: `TraverseObjectGraphRecursively`

PR #1874, opened 2020-05-19
Added a setting for running Hangfire in SqlServer mode (instead of MemoryStorage)

Jounik53:
Changed the way to launch Hangfire. By default SqlServer mode should be used instead of MemoryStorage


lnetrebskii:
Review task https://virtocommerce.atlassian.net/browse/VP-2394 has been created

PR #1877, opened 2020-05-21
VP-2440: Add CI pipeline

lnetrebskii:
### Problem
1. External contributors don't have access to CI results for details.
2. Experiment with using GitHub Actions instead of Jenkins.

### Solution
Create a GitHub Actions CI pipeline with essential steps to verify Pull Requsts.

### Proposed of changes
A new file with a pipeline has been implemented.
.github/workflows/main.yml

N/A

- [X] Check methods and variable namings - it should be self descriptive, no typos
- [X] Check you did not introduce breaking changes in API and public models/services.
- [X] Respect extensibility - https://community.virtocommerce.com/t/extensibility-basics-the-domain-model-and-persistence-layer-extension/141
- [X] Follow [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself) and [SOLID](https://en.wikipedia.org/wiki/SOLID) principles
- [X] For unit tests - follow ms best practices: https://docs.microsoft.com/en-us/dotnet/core/testing/unit-testing-best-practices
- [X] Consolidate solution dependencies in case you are using newer version, update VC module dependencies in module.manifest. Do not upgrade 3rd party packages that are shipped with the platform with the version newer than in the platform.
- [X] Check code style conventions - https://github.com/VirtoCommerce/styleguide/blob/master/csharp.md
- [X] Check PR have a concise and descriptive title, follow [git commit message rules](https://github.com/VirtoCommerce/styleguide/blob/master/gitcommits.md)

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -25,3 +25,6 @@ jobs:
 
     - name: Build
       run: vc-build Compile
+
+    - name: Unit Tests
+      run: vc-build Test -skip Restore+Compile
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -20,12 +20,8 @@ jobs:
     # Checks-out your repository under $GITHUB_WORKSPACE, so your job can access it
     - uses: actions/checkout@v2
 
-    # Runs a single command using the runners shell
     - name: Install VirtoCommerce.GlobalTool
       run: dotnet tool install VirtoCommerce.GlobalTool --version 3.0.0-beta0006  -g
 
-    # Runs a set of commands using the runners shell
-    - name: Run a multi-line script
-      run: |
-        echo Add other actions to build,
-        echo test, and deploy your project.
+    - name: Build
+      run: vc-build Compile
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -22,7 +22,7 @@ jobs:
 
     # Runs a single command using the runners shell
     - name: Install VirtoCommerce.GlobalTool
-      run: dotnet tool install VirtoCommerce.GlobalTool  -g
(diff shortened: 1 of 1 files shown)


.github/workflows/main.yml
```diff
@@ -0,0 +1,33 @@
...
+    # The type of runner that the job will run on
+    runs-on: windows-latest
+
+    # Steps represent a sequence of tasks that will be executed as part of the job
+    steps:
+    # Checks-out your repository under $GITHUB_WORKSPACE, so your job can access it
+    - uses: actions/checkout@v2
+
+    - name: Install VirtoCommerce.GlobalTool
+      run: dotnet tool install --global VirtoCommerce.GlobalTool --version 3.0.0-beta0010
```
yecli, line 24: Should we install the specific version? 
--version 3.0.0-beta0010
I think to avoid changes with new Global tool versions it is better to use the latest.

lnetrebskii: It doesn't work because it's still in a pre-release.

yecli: I see

.github/workflows/main.yml
```diff
@@ -0,0 +1,33 @@
...
+
+    - name: Build
+      run: vc-build Compile
+
+    - name: Unit Tests
+      run: vc-build Test -skip Restore+Compile
+
+    - name: Build Package
+      run: vc-build Compress -skip Clean+Restore+Compile+Test
```
tatarincev, line 33: you can leave only the `vc-build Compress `  it will execute all targets anyway.

lnetrebskii: I know. I split it into independent steps to visualize progress.

PR #1878, opened 2020-05-21
VP-2395: Add deploy to Azure button to readme

yecli:


yecli:
Need to update Azure deployment scripts. Will create a separate task on this.

PR #1880, opened 2020-05-23
VP-2408 elimination memory leaks

tatarincev:


src/VirtoCommerce.Platform.Core/Common/AsyncLock.cs
```diff
@@ -56,41 +56,54 @@ public void Release()
...
+                m_toRelease = toRelease;
+                m_postReleaseAction = postReleaseAction;
+            }
 
             public void Dispose()
             {
                 if (m_toRelease != null)
+                {
                     m_toRelease.m_semaphore.Release();
+                    if (m_toRelease != null)
```
yecli, line 102: Should be
```
if (m_postReleaseAction != null)
```
?

src/VirtoCommerce.Platform.Core/Common/AsyncLock.cs
```diff
@@ -56,41 +56,54 @@ public void Release()
     // http://blogs.msdn.com/b/pfxteam/archive/2012/02/12/10266988.aspx
```
akak1977, line 56: This reference is expired. But we can use https://devblogs.microsoft.com/pfxteam/building-async-coordination-primitives-part-6-asynclock/. Does it sense?

PR #1882, opened 2020-05-25
VP-2309: [BUG] Using + in User name

pushnitsa:
### Problem
When application runs with IIS. There is unable to make request with symbol "+" into the route

### Solution
Here solution and explanation
https://stackoverflow.com/questions/1453218/is-enabling-double-escaping-dangerous

### Proposed of changes
Allow double escaping

- [x] For unit tests - follow ms best practices: https://docs.microsoft.com/en-us/dotnet/core/testing/unit-testing-best-practices


src/VirtoCommerce.Platform.Web/web.config
```diff
@@ -16,9 +16,9 @@
       </environmentVariables>
     </aspNetCore>
     <security>
-      <requestFiltering>
+      <requestFiltering allowDoubleEscaping="true">
```
tatarincev, line 19: But this will work only for "in-process" hosting mode, can you check how it will work for "outofprocess" mode?
And I think these changes can significantly affect all API and required full API regress tests. I think before applying this fix we need to get more know-how this option `allowDoubleEscaping`  works.

PR #1886, opened 2020-05-25
#VP-2577 fix menu style

kostyrin:


src/VirtoCommerce.Platform.Web/wwwroot/css/themes/main/sass/modules/_menu.sass
```diff
@@ -36,6 +36,7 @@
     z-index: 50
     align-items: center
     display: flex
+    background: #161d25
```
lnetrebskii, line 39: There is a constant $headerBg with the same value. Please reuse it.

PR #1887, opened 2020-05-26
#VP-2526 hangfire migration from v2

kostyrin:


src/VirtoCommerce.Platform.Data/Migrations/20000000000000_UpdateHangfireV2.cs
```diff
@@ -9,6 +9,7 @@ protected override void Up(MigrationBuilder migrationBuilder)
             migrationBuilder.Sql(@"IF ((EXISTS (SELECT * FROM INFORMATION_SCHEMA.TABLES WHERE TABLE_NAME = '__MigrationHistory')) AND 
 	                (EXISTS (SELECT * FROM INFORMATION_SCHEMA.TABLES WHERE TABLE_SCHEMA = 'HangFire')))
                 IF (EXISTS (SELECT * FROM __MigrationHistory WHERE ContextKey = 'VirtoCommerce.Platform.Data.Repositories.Migrations.Configuration'))
+                    AND (SELECT TOP 1 Version FROM [HangFire].[Schema]) < 7
```
akak1977, line 12: I think we should always drop old hangfire scheme during v2->v3 upgrade. 
Because there will be jobs with obsolete codebase.
It seems the best way is always start with empty hangfire DB.

kostyrin: have deleted it but please read
https://virtocommerce.atlassian.net/browse/VP-2526?focusedCommentId=22352

yecli:
Not needed anymore, solved another way.

PR #1888, opened 2020-05-27
VP-2526 Hangfire configuration

tatarincev:


lnetrebskii:
Review task https://virtocommerce.atlassian.net/browse/VP-2610 has been created

src/VirtoCommerce.Platform.Data/Infrastructure/DbContextUnitOfWork.cs
```diff
@@ -27,21 +27,12 @@ public DbContextUnitOfWork(DbContext context)
...
-            try
-            {
-                var result = await DbContext.SaveChangesAsync();
-                return result;
-            }
-            catch (Exception ex)
-            {
-                throw new PlatformException(ex.ExpandExceptionMessage());
-            }
+            return DbContext.SaveChangesAsync();
```
yecli, line 35: Why `await` is removed here?
Now even if CommitAsync will be awaited, it will not wait for SaveChangesAsync completion, isn't it?

yecli: My wrong, SaveChangesAsync will be awaited in that case.

PR #1890, opened 2020-05-27
VP-2538: Fix error on lock out

yecli:
Added cache expiration (otherwise locked out user could behave like the normal one for some time on the storefront)


src/VirtoCommerce.Platform.Web/Security/CustomUserManager.cs
```diff
@@ -216,7 +216,25 @@ public override async Task<IdentityResult> CreateAsync(ApplicationUser user)
             return result;
         }
 
+        public override async Task<IdentityResult> SetLockoutEndDateAsync(ApplicationUser user, DateTimeOffset? lockoutEnd)
```
tatarincev, line 219: I think we don't need this overriding, looking to original  `SetLockoutEndDateAsync` implementation Asp.NET Identity https://github.com/dotnet/aspnetcore/blob/master/src/Identity/Extensions.Core/src/UserManager.cs#L2007 you can see what the `UpdateUserAsync` method is called. Because we have an overridden version of `UpdateUserAsync` method all these logic will be called anyway. There must be another reason for this problem.

yecli: The problem here is that we are overriding [UpdateAsync](https://github.com/VirtoCommerce/vc-platform/blob/dev/src/VirtoCommerce.Platform.Web/Security/CustomUserManager.cs#L141), not [UpdateUserAsync](https://github.com/dotnet/aspnetcore/blob/master/src/Identity/Extensions.Core/src/UserManager.cs#L2007) you mentioned.
So probably we need to override `UpdateUserAsync` function, which is called on all changes operations and is also [called](https://github.com/dotnet/aspnetcore/blob/master/src/Identity/Extensions.Core/src/UserManager.cs#L503) from built-in `UpdateAsync`, right?

tatarincev: Yes, Indeed

PR #1902, opened 2020-06-02
VP-1681: Platform > Modules deselected when I select modules bundle and versa vers

pushnitsa:


src/VirtoCommerce.Platform.Web/wwwroot/js/app/modularity/blades/modules-list.js
```diff
@@ -1,4 +1,4 @@
-﻿angular.module('platformWebApp')
+angular.module('platformWebApp')
```
yecli, line 1: Seems this is redundant change, could be reverted

PR #1905, opened 2020-06-03
VP-2593: Add UpdateUserAsync override

yecli:
- Now every user update operation will trigger events and cache expiration;
- Drawback - updating user with new/deleted roles will result in multiple events;
- Refactoring;


src/VirtoCommerce.Platform.Web/Security/CustomUserManager.cs
```diff
@@ -138,57 +138,65 @@ public override async Task<IdentityResult> DeleteAsync(ApplicationUser user)
...
                     //Remove
                     foreach (var removeRole in targetRoles.Except(sourceRoles))
                     {
-                        await RemoveFromRoleAsync(existUser, removeRole);
+                        await RemoveFromRoleAsync(user, removeRole);
                     }
                 }
-                SecurityCacheRegion.ExpireUser(existUser);
+
+                SecurityCacheRegion.ExpireUser(user);
```
tatarincev, line 197: this line is redundant. The cache will be invalidated on `UpdateUserAsync` call

PR #1911, opened 2020-06-08
VP-2780: Add SignalR scalability configuration.

trueboroda:
SignalR scalability can be configured with "SignalR" section of  platform configuration.
To enable scalability set "ScalabilityProvider" parameter to available value. 
You can set one of two providers, "AzureSignalRService" or "RedisBackplane".  
The first way also requires to configure connection string in "AzureSignalRService" section. 
If RedisBackplane mode is set, the common Redis connection string will be used. You can set the  Redis channel name via "RedisBackplane" section.
```
 "SignalR": {        
        "ScalabilityProvider": null,
        "AzureSignalRService": {              
            "ConnectionString": "AzureSignalRServiceConnectionString"                               
        },
        "RedisBackplane": {
            "ChannelName": "VirtoCommerceChannel"            
        }
    }
```


yecli:
The only thing I would suggest is to add `RedisBackplaneOptions` with the ability to specify ChannelName.
It could be set in the connection string, but as we are already using this connection string, we may want to specify different channels for different cases.

PR #1914, opened 2020-06-10
VP-2904: Extend ChangeLogSearchCriteria with OperationType property

trueboroda:
Due to https://virtocommerce.atlassian.net/browse/VP-2761


src/VirtoCommerce.Platform.Core/ChangeLog/ChangeLogSearchCriteria.cs
```diff
@@ -5,6 +5,7 @@ namespace VirtoCommerce.Platform.Core.ChangeLog
 {
     public class ChangeLogSearchCriteria : SearchCriteriaBase
     {
+        public EntryState? OperationType { get; set; }
```
yecli, line 8: I think it is better to add EntryState[] property and add conditions only if it is not empty.

PR #1926, opened 2020-06-22
VP-3207: Synchronize push notifications history between multiple platform instances

trueboroda:
In addition to upgrade binaries, you also will need to set `PushNotificationHubConnectionString` to `http(s)://{admin_url}/pushNotificationHub` where `admin_url` is the url of admin you configure, for example:
`https://dev-demovc3-admin.azurewebsites.net/pushNotificationHub` for first instance of DEV v3 admin and
`https://dev-demovc3-admin-second.azurewebsites.net/pushNotificationHub` for second instance of DEV v3 admin


src/VirtoCommerce.Platform.Core/JsonConverters/PolymorphJsonConverter.cs
```diff
@@ -4,14 +4,21 @@
...
+using VirtoCommerce.Platform.Core.PushNotifications;
 using VirtoCommerce.Platform.Core.Security;
 using VirtoCommerce.Platform.Core.Settings;
 
 namespace VirtoCommerce.Platform.Core.JsonConverters
 {
     public class PolymorphJsonConverter : JsonConverter
     {
-        private static readonly Type[] _knowTypes = { typeof(ObjectSettingEntry), typeof(DynamicProperty), typeof(ApplicationUser), typeof(Role), typeof(PermissionScope) };
+        private static readonly Type[] _knowTypes = { typeof(ObjectSettingEntry), typeof(DynamicProperty), typeof(ApplicationUser), typeof(Role), typeof(PermissionScope), typeof(PushNotification) };
```
tatarincev, line 15: I think it would be better to reject of using strongly typed `PuschNotification` for inter-process communication, and define the special `DistributedPushNotification` type instead, that will be derived from `PushNotification` and contains a `Dictionary<string, object> Properties` that will be serialized and deserialized from/into Json absolutely transparently for non-strongly-type clients such as virto manager SPA.
https://stackoverflow.com/questions/1207731/how-can-i-deserialize-json-to-a-simple-dictionarystring-string-in-asp-net

src/VirtoCommerce.Platform.Core/SignalR/SignalROptions.cs
```diff
@@ -0,0 +1,11 @@
+namespace VirtoCommerce.Platform.Core.SignalR
+{
+    public class SignalROptions
```
tatarincev, line 3: It is absolutely bad practice to mixed the two different infrastructure per single POCO options type.
You need to define the two different options that can be defined into their own namespaces and register them independently based on the mode setting. e.g https://github.com/VirtoCommerce/vc-platform/blob/master/src/VirtoCommerce.Platform.Web/Startup.cs#L384

src/VirtoCommerce.Platform.Web/SignalR/ServiceCollectionExtensions.cs
```diff
@@ -0,0 +1,58 @@
...
+            {
+                throw new InvalidOperationException("SignalR:AzureSignalRService:ConnectionString  must be set");
+            }
+            builder.AddAzureSignalR(options =>
+            {
+                options.Endpoints = new [] { new ServiceEndpoint(azureSignalRConnectionString) };
+            });
+        }
+
+        public static void AddRedisBackplane(this ISignalRServerBuilder builder, IConfiguration configuration)
```
tatarincev, line 49: This registration can be moved from here to the Redis `serviceCollectionExtensions`.

src/VirtoCommerce.Platform.Web/appsettings.json
```diff
@@ -152,5 +151,10 @@
         "RedisBackplane": {
             "ChannelName": "VirtoCommerceChannel"
         }
-    }
+    },
+    "PushNotifications": {
```
tatarincev, line 155: I'm not sure what we need the special `PushNotifications` section in config, the one SignalR will be enough because we do not use SignalR for other logic than PushNotification. The setting   `HubUrl` also can be part of AzureSignalR settings.

I suggest using the following config structure
```JSON
 "SignalR": {
        //Possible values: RedisBackplane (default) | AzureSignalRService
        //RedisBackplane will be activated only when RedisConnectionString is set otherwise no any SignalR scaling options will be used 
        "ScalabilityProvider": "RedisBackplane",
        "AzureSignalRService": {
            "HubUrl": "http://localhost:10645/pushNotificationHub"
            "ConnectionString": "Endpoint=https://{your service name}.service.signalr.net;AccessKey={your access key};Version=1.0;"
        },
        "RedisBackplane": {
            "ChannelName": "VirtoCommerceChannel"
        }
    },
```

asvishnyakov: 1. `HubUrl` is specific for push notifications and not related ti SignalR confguration. SignalR can be used for other purposes too.
2. `AzureSignalRService` options will be loaded only if Azure SignalR service is used

src/VirtoCommerce.Platform.Core/JsonConverters/PolymorphJsonConverter.cs
asvishnyakov: Built-in type name handling using now

src/VirtoCommerce.Platform.Core/SignalR/SignalROptions.cs
asvishnyakov: Fixed

src/VirtoCommerce.Platform.Web/SignalR/ServiceCollectionExtensions.cs
asvishnyakov: Fixed

asvishnyakov:
It's impossible to somehow get and use app URL to connect to push notification hub:
1. No API for directly connection to Signal R server - only HTTP client is available
2. We can't use `HttpContext` to get URL from `Request` - it's unavailable at `BackgroundService` execution time
3. We can't use `IServerFeature` to get and use current app port (i.e. connect to `127.0.0.1:RANDOM_KESTREL_PORT`) - it's unavailable at startup time too

asvishnyakov:
JsonSerializationSettings:
1. I introduced some changes to copy it, because SignalR use it's own options, not the same object as we configure for MVC
2. `TypeNameHandling` and `TypeNameAssemblyFormatHandling` is used for push notification strong typification. They are specified for SignalR & push notifications only

PR #1927, opened 2020-06-23
VP-3266: Disable & reset cache if Redis is temporary unavailable

asvishnyakov:
### Problem
If Redis is unavailable (for example, at Redis reboot time) running multiple instances of platform isn't safe for cache: without Redis they will be out of sync and have different values because cache of other instances will not be invalidated after changes in one of them.

### Solution
Disable & reset cache if Redis temporary unavailable, then enable (if it was) and reset it again after connection established.

### Proposed of changes
1. Update platform version and merge to master as part of hotfix strategy
2. Create `GlobalCacheRegion` to have ability reset all cache entries
3. Allow disable & reset cache
3.1. Extend `IPlatformMemoryCache` with `EnableCache`, `AbsoluteExpiration` and `SlidingExpiration`
3.2. Make `EnableCache`, `AbsoluteExpiration` and `SlidingExpiration` virtual for extensibility and allow to set value without changing options (we need to keep them initial values to restore later)
3.3. Add change token from `GlobalCacheRegion` to every cache entry
4. Update `RedisPlatformMemoryCache`
4.1. Add failed connection and connection restore loggin, AI events, exception tracking.
4.2. **Disable & reset cache if Redis temporary unavailable, then enable (if it was) and reset it again after connection established.**
4.3. Change subscription flags back to `CommandFlags.FireAndForget` as it was in Platform v2. Platform doesn't need to wait Redis response from commands
4.4. Add subscription event to AI
4.5. Make OnMessage extensible
4.6. Remove Polly because Redis manage retry policy itself
4.7. Mark `BusRetryCount` obsolete as no longer used
5. It isn't part of PR, but it should be recommended to use at least `STANDARD` plan for Redis and decrease connection checking timeouts at least on production: `keepAlive=10,configCheckSeconds=10`

### DIFF
--- a/Directory.Build.Props
+++ b/Directory.Build.Props
@@ -7,7 +7,7 @@
     <Authors>VirtoCommerce</Authors>
   </PropertyGroup>
    <PropertyGroup>
-    <VersionPrefix>3.2.0</VersionPrefix>
+    <VersionPrefix>3.2.1</VersionPrefix>
     <VersionSuffix></VersionSuffix>
     <VersionSuffix Condition=" '$(VersionSuffix)' != '' AND '$(BuildNumber)' != '' ">$(VersionSuffix)-$(BuildNumber)</VersionSuffix>
   </PropertyGroup>
--- a/src/VirtoCommerce.Platform.Caching/PlatformMemoryCache.cs
+++ b/src/VirtoCommerce.Platform.Caching/PlatformMemoryCache.cs
@@ -9,9 +9,6 @@ namespace VirtoCommerce.Platform.Caching
     public class PlatformMemoryCache : IPlatformMemoryCache
     {
         private readonly IMemoryCache _memoryCache;
-        private bool _cacheEnabled;
-        private TimeSpan? _absoluteExpiration;
-        private TimeSpan? _slidingExpiration;
         private bool _disposed;
         private readonly ILogger _log;
 
@@ -19,9 +16,9 @@ namespace VirtoCommerce.Platform.Caching
         {
             _memoryCache = memoryCache;
             var cachingOptions = options.Value;
-            _cacheEnabled = cachingOptions.CacheEnabled;
-            _absoluteExpiration = cachingOptions.CacheAbsoluteExpiration;
-            _slidingExpiration = cachingOptions.CacheSlidingExpiration;
+            CacheEnabled = cachingOptions.CacheEnabled;
+            AbsoluteExpiration = cachingOptions.CacheAbsoluteExpiration;
+            SlidingExpiration = cachingOptions.CacheSlidingExpiration;
             _log = log;
         }
         
@@ -48,23 +45,11 @@ namespace VirtoCommerce.Platform.Caching
         }
 
 
-        public virtual bool CacheEnabled
-        {
-            get { return _cacheEnabled ; }
-            set { _cacheEnabled = value; }
-        }
+        protected bool CacheEnabled { get; set; }
 
-        public virtual TimeSpan? AbsoluteExpiration
-        {
-            get { return _absoluteExpiration; }
-            set { _absoluteExpiration = value; }
-        }
+        protected TimeSpan? AbsoluteExpiration { get; set; }
 
-        public virtual TimeSpan? SlidingExpiration
-        {
-            get { return _slidingExpiration; }
-            set { _slidingExpiration = value; }
-        }
+        protected TimeSpan? SlidingExpiration { get; set; }
 
 
         public virtual MemoryCacheEntryOptions GetDefaultCacheEntryOptions()
--- a/src/VirtoCommerce.Platform.Caching/PlatformMemoryCache.cs
+++ b/src/VirtoCommerce.Platform.Caching/PlatformMemoryCache.cs
@@ -46,19 +46,19 @@ namespace VirtoCommerce.Platform.Caching
         }
 
 
-        public virtual bool CacheEnabled
+        protected virtual bool CacheEnabled
         {
             get { return _cacheEnabled ?? _cachingOptions.CacheEnabled; }
             set { _cacheEnabled = _cachingOptions.CacheEnabled == value ? (bool?) null : value; }
         }
 
-        public virtual TimeSpan? AbsoluteExpiration
+        protected virtual TimeSpan? AbsoluteExpiration
         {
             get { return _absoluteExpiration ?? _cachingOptions.CacheAbsoluteExpiration; }
             set { _absoluteExpiration = _cachingOptions.CacheAbsoluteExpiration == value ? null : value; }
         }
 
-        public virtual TimeSpan? SlidingExpiration
+        protected virtual TimeSpan? SlidingExpiration
         {
             get { return _slidingExpiration ?? _cachingOptions.CacheSlidingExpiration; }
             set { _slidingExpiration = _cachingOptions.CacheSlidingExpiration == value ? null : value; }
--- a/src/VirtoCommerce.Platform.Caching/PlatformMemoryCache.cs
+++ b/src/VirtoCommerce.Platform.Caching/PlatformMemoryCache.cs
@@ -46,19 +46,19 @@ namespace VirtoCommerce.Platform.Caching
         }
 
 
-        protected virtual bool CacheEnabled
+        public virtual bool CacheEnabled
         {
             get { return _cacheEnabled ?? _cachingOptions.CacheEnabled; }
             set { _cacheEnabled = _cachingOptions.CacheEnabled == value ? (bool?) null : value; }
         }
 
-        protected virtual TimeSpan? AbsoluteExpiration
+        public virtual TimeSpan? AbsoluteExpiration
         {
             get { return _absoluteExpiration ?? _cachingOptions.CacheAbsoluteExpiration; }
             set { _absoluteExpiration = _cachingOptions.CacheAbsoluteExpiration == value ? null : value; }
         }
 
-        protected virtual TimeSpan? SlidingExpiration
+        public virtual TimeSpan? SlidingExpiration
         {
             get { return _slidingExpiration ?? _cachingOptions.CacheSlidingExpiration; }
             set { _slidingExpiration = _cachingOptions.CacheSlidingExpiration == value ? null : value; }
--- a/src/VirtoCommerce.Platform.Caching/PlatformMemoryCache.cs
+++ b/src/VirtoCommerce.Platform.Caching/PlatformMemoryCache.cs
@@ -8,9 +8,8 @@ namespace VirtoCommerce.Platform.Caching
 {
     public class PlatformMemoryCache : IPlatformMemoryCache
     {
-        private readonly CachingOptions _cachingOptions;
         private readonly IMemoryCache _memoryCache;
-        private bool? _cacheEnabled;
+        private bool _cacheEnabled;
         private TimeSpan? _absoluteExpiration;
         private TimeSpan? _slidingExpiration;
         private bool _disposed;
@@ -19,7 +18,10 @@ namespace VirtoCommerce.Platform.Caching
         public PlatformMemoryCache(IMemoryCache memoryCache, IOptions<CachingOptions> options, ILogger<PlatformMemoryCache> log)
         {
             _memoryCache = memoryCache;
-            _cachingOptions = options.Value;
+            var cachingOptions = options.Value;
+            _cacheEnabled = cachingOptions.CacheEnabled;
+            _absoluteExpiration = cachingOptions.CacheAbsoluteExpiration;
+            _slidingExpiration = cachingOptions.CacheSlidingExpiration;
             _log = log;
         }
         
@@ -48,20 +50,20 @@ namespace VirtoCommerce.Platform.Caching
 
         public virtual bool CacheEnabled
         {
-            get { return _cacheEnabled ?? _cachingOptions.CacheEnabled; }
-            set { _cacheEnabled = _cachingOptions.CacheEnabled == value ? (bool?) null : value; }
+            get { return _cacheEnabled ; }
+            set { _cacheEnabled = value; }
         }
 
         public virtual TimeSpan? AbsoluteExpiration
         {
-            get { return _absoluteExpiration ?? _cachingOptions.CacheAbsoluteExpiration; }
-            set { _absoluteExpiration = _cachingOptions.CacheAbsoluteExpiration == value ? null : value; }
+            get { return _absoluteExpiration; }
+            set { _absoluteExpiration = value; }
         }
 
         public virtual TimeSpan? SlidingExpiration
         {
-            get { return _slidingExpiration ?? _cachingOptions.CacheSlidingExpiration; }
-            set { _slidingExpiration = _cachingOptions.CacheSlidingExpiration == value ? null : value; }
+            get { return _slidingExpiration; }
+            set { _slidingExpiration = value; }
         }
 
 
--- a/src/VirtoCommerce.Platform.Caching/PlatformMemoryCache.cs
+++ b/src/VirtoCommerce.Platform.Caching/PlatformMemoryCache.cs
@@ -10,6 +10,9 @@ namespace VirtoCommerce.Platform.Caching
     {
         private readonly CachingOptions _cachingOptions;
         private readonly IMemoryCache _memoryCache;
+        private bool? _cacheEnabled;
+        private TimeSpan? _absoluteExpiration;
+        private TimeSpan? _slidingExpiration;
         private bool _disposed;
         private readonly ILogger _log;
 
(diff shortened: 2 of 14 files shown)


src/VirtoCommerce.Platform.Caching/Redis/RedisPlatformMemoryCache.cs
```diff
@@ -60,9 +103,6 @@ protected override void EvictionCallback(object key, object value, EvictionReaso
         {
             _log.LogInformation($"{nameof(RedisPlatformMemoryCache)}: channel[{_redisCachingOptions.ChannelName }] sending a message with key:{key} from instance:{ _cacheId } to all subscribers");
 
-            var message = new RedisCachingMessage { Id = _cacheId, CacheKeys = new[] { key } };
```
tatarincev, line 63: Why you deleted this line? How the platform will inform the other instances about cache key eviction?

src/VirtoCommerce.Platform.Caching/Redis/RedisPlatformMemoryCache.cs
```diff
@@ -1,47 +1,90 @@
...
+            {
+                {"channelName", _redisCachingOptions.ChannelName},
+                {"cacheId", _cacheId},
+                {"endpoint", e.EndPoint.ToString()},
+                {"failureType", e.FailureType.ToString()}
+            });
 
-            _retryPolicy = Policy.Handle<Exception>().WaitAndRetry(
-                _redisCachingOptions.BusRetryCount, retryAttempt => TimeSpan.FromSeconds(Math.Pow(2, retryAttempt - 1)));
+            CacheEnabled = false;
```
tatarincev, line 69: Please add some comments why you need to do this

src/VirtoCommerce.Platform.Caching/PlatformMemoryCache.cs
```diff
@@ -43,12 +46,26 @@ public virtual void Remove(object key)
         }
 
 
-        protected bool CacheEnabled => _cachingOptions.CacheEnabled;
-        protected TimeSpan? AbsoluteExpiration => _cachingOptions.CacheAbsoluteExpiration;
-        protected TimeSpan? SlidingExpiration => _cachingOptions.CacheSlidingExpiration;
+        public virtual bool CacheEnabled
```
tatarincev, line 49: Why have you changed accessibility for these properties?

src/VirtoCommerce.Platform.Caching/Redis/RedisPlatformMemoryCache.cs
```diff
@@ -1,47 +1,90 @@
...
 using VirtoCommerce.Platform.Caching;
+using VirtoCommerce.Platform.Core.Caching;
 using VirtoCommerce.Platform.Core.Common;
 
 namespace VirtoCommerce.Platform.Redis
 {
     public class RedisPlatformMemoryCache : PlatformMemoryCache
     {
         private readonly ISubscriber _bus;
+        private readonly IMemoryCache _memoryCache;
```
t13ka, line 18: it is not used. should be removed

src/VirtoCommerce.Platform.Caching/ServiceCollectionExtensions.cs
```diff
@@ -8,20 +14,22 @@ namespace VirtoCommerce.Platform.Caching
...
 
-            if (!string.IsNullOrEmpty(redisConnectionString))
+            if (!string.IsNullOrEmpty(_redisConnectionString))
             {
                 services.AddOptions<RedisCachingOptions>().Bind(configuration.GetSection("Caching:Redis")).ValidateDataAnnotations();
 
-                var redis = ConnectionMultiplexer.Connect(redisConnectionString);
+                var redis = ConnectionMultiplexer.Connect(_redisConnectionString);
+                services.AddSingleton(redis);
                 services.AddSingleton(redis.GetSubscriber());
```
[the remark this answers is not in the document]
asvishnyakov: It's breaking change

Directory.Build.Props
```diff
@@ -7,7 +7,7 @@
     <Authors>VirtoCommerce</Authors>
   </PropertyGroup>
    <PropertyGroup>
-    <VersionPrefix>3.2.0</VersionPrefix>
+    <VersionPrefix>3.2.1</VersionPrefix>
```
tatarincev, line 10: You mustn't increment platform version by your own.

src/VirtoCommerce.Platform.Caching/Redis/RedisPlatformMemoryCache.cs
```diff
@@ -1,47 +1,90 @@
...
+        private readonly CachingOptions _cachingOptions;
         private readonly RedisCachingOptions _redisCachingOptions;
         private readonly ILogger _log;
-        private readonly RetryPolicy _retryPolicy;
+        private readonly TelemetryClient _telemetryClient;
```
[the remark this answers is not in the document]
asvishnyakov: We can, but I'll prefer directly use `TelemetryClient` because it allow you to provide more information and more accurately manage data in AI

src/VirtoCommerce.Platform.Caching/Redis/RedisPlatformMemoryCache.cs
```diff
@@ -1,47 +1,90 @@
...
 
             _log.LogInformation($"{nameof(RedisPlatformMemoryCache)}: subscribe to channel {_redisCachingOptions.ChannelName } current instance:{ _cacheId }");
+            _telemetryClient.TrackEvent("RedisSubscribed", new Dictionary<string, string>
+            {
+                {"channelName", _redisCachingOptions.ChannelName},
+                {"cacheId", _cacheId}
+            });
+        }
+
+        protected virtual void OnConnectionFailed(object? sender, ConnectionFailedEventArgs e)
```
t13ka: for what reason object type marked with a question mark at the end? I guess there is no difference so I suggest you remove it

Directory.Build.Props
asvishnyakov: See PR description: 
1. Update platform version and merge to master as part of hotfix strategy

Ask @lnetrebskii for more information

src/VirtoCommerce.Platform.Caching/Redis/RedisPlatformMemoryCache.cs
```diff
@@ -1,47 +1,90 @@
...
 using StackExchange.Redis;
 using VirtoCommerce.Platform.Caching;
+using VirtoCommerce.Platform.Core.Caching;
 using VirtoCommerce.Platform.Core.Common;
 
 namespace VirtoCommerce.Platform.Redis
 {
     public class RedisPlatformMemoryCache : PlatformMemoryCache
     {
         private readonly ISubscriber _bus;
```
t13ka, line 17: _bus can be converted to local var

src/VirtoCommerce.Platform.Caching/PlatformMemoryCache.cs
asvishnyakov: For extensibility reasons I added these properties to `IPlatformMemoryCache`. Interface members can't be implemented as protected (except implicit implementation aka `void IPlatformMemoryCache.CacheEnabled`). They also should be public for same extensibility reasons - otherwise you can't disable cache without making your own `PlatformMemoryCache` ancestor, which isn't best solution especially because of existing `RedisPlatformMemoryCache`

src/VirtoCommerce.Platform.Caching/Redis/RedisPlatformMemoryCache.cs
asvishnyakov: @tatarincev Done, review please

src/VirtoCommerce.Platform.Caching/Redis/RedisPlatformMemoryCache.cs
```diff
@@ -1,47 +1,90 @@
...
         private readonly RedisCachingOptions _redisCachingOptions;
         private readonly ILogger _log;
-        private readonly RetryPolicy _retryPolicy;
+        private readonly TelemetryClient _telemetryClient;
 
         private static readonly string _cacheId = Guid.NewGuid().ToString("N");
         
 
         public RedisPlatformMemoryCache(IMemoryCache memoryCache, IOptions<CachingOptions> options
+            , ConnectionMultiplexer connection
```
t13ka, line 28: Can we pass it here as Interface?

src/VirtoCommerce.Platform.Caching/Redis/RedisPlatformMemoryCache.cs
```diff
@@ -59,9 +110,9 @@ private void OnMessage(RedisChannel channel, RedisValue redisValue)
         protected override void EvictionCallback(object key, object value, EvictionReason reason, object state)
         {
             _log.LogInformation($"{nameof(RedisPlatformMemoryCache)}: channel[{_redisCachingOptions.ChannelName }] sending a message with key:{key} from instance:{ _cacheId } to all subscribers");
-
+            
             var message = new RedisCachingMessage { Id = _cacheId, CacheKeys = new[] { key } };
-            _retryPolicy.Execute(() => _bus.Publish(_redisCachingOptions.ChannelName, JsonConvert.SerializeObject(message)));
+            _bus.Publish(_redisCachingOptions.ChannelName, JsonConvert.SerializeObject(message));
```
tatarincev, line 115: you forgot to add FireAndForget options

src/VirtoCommerce.Platform.Caching/PlatformMemoryCache.cs
tatarincev: you don't need to extend the entire interface `IPlatformMemoryCache` due to these extends the responsibility and absolutely not required for this task. Just call protected properties in the descendant.

src/VirtoCommerce.Platform.Caching/Redis/RedisPlatformMemoryCache.cs
```diff
@@ -1,47 +1,90 @@
...
+                {"failureType", e.FailureType.ToString()}
+            });
 
-            _retryPolicy = Policy.Handle<Exception>().WaitAndRetry(
-                _redisCachingOptions.BusRetryCount, retryAttempt => TimeSpan.FromSeconds(Math.Pow(2, retryAttempt - 1)));
+            CacheEnabled = false;
+            GlobalCacheRegion.ExpireRegion();
+        }
+
+        protected virtual void OnConnectionRestored(object? sender, ConnectionFailedEventArgs e)
```
t13ka: What is the goal to do this and other methods as virtual and protected? Is there is a small chance of overriding in the spectral future? :)
We have an important logic in methods, these logics should be tested by mocks verification, eg:
in method OnConnectionFailed we have cache disabling, this important part can be tested by unit-testing.
Also, we can verify the methods invocations eg: TrackException, TrackEvent

src/VirtoCommerce.Platform.Caching/Redis/RedisPlatformMemoryCache.cs
asvishnyakov: Reverted back

src/VirtoCommerce.Platform.Caching/PlatformMemoryCache.cs
asvishnyakov: `IPlatformMemoryCache` inherit `IMemoryCache` which is responsible for manipulations with cache (get, add, remove cache entries). It also provide cache options for same cases - cache disability & expirations. So it's ok to add them here. I'm just trying to make minor improvements to prevent problems with extensibility later.

But as you wish.

@OlegoO And you saying "make small improvements if you see problems". How?)

src/VirtoCommerce.Platform.Caching/Redis/RedisPlatformMemoryCache.cs
```diff
@@ -1,47 +1,98 @@
...
+            _telemetryClient = telemetryClient;
             _bus = bus;
 
+            _cachingOptions = cachingOptions.Value;
             _redisCachingOptions = redisCachingOptions.Value;
-            _bus.Unsubscribe(_redisCachingOptions.ChannelName);
-            _bus.Subscribe(_redisCachingOptions.ChannelName, OnMessage);
+
+            connection.ConnectionFailed += OnConnectionFailed;
+            connection.ConnectionRestored += OnConnectionRestored;
```
t13ka, line 45: subscription without unsubscription is the potential dangerous reason for memory leaking

src/VirtoCommerce.Platform.Caching/Redis/RedisPlatformMemoryCache.cs
asvishnyakov: Fxed

src/VirtoCommerce.Platform.Caching/Redis/RedisPlatformMemoryCache.cs
asvishnyakov: Fixed

src/VirtoCommerce.Platform.Caching/Redis/RedisPlatformMemoryCache.cs
asvishnyakov: Fixed

src/VirtoCommerce.Platform.Caching/Redis/RedisPlatformMemoryCache.cs
asvishnyakov: Not after code changes, it's used in `EvictionCallback`

src/VirtoCommerce.Platform.Caching/Redis/RedisPlatformMemoryCache.cs
```diff
@@ -1,47 +1,98 @@
...
+            });
 
-            _retryPolicy = Policy.Handle<Exception>().WaitAndRetry(
-                _redisCachingOptions.BusRetryCount, retryAttempt => TimeSpan.FromSeconds(Math.Pow(2, retryAttempt - 1)));
+            // If we have no connection to Redis, we can't invalidate cache on another platform instances,
+            // so the better idea is to disable cache at all for data consistence
+            CacheEnabled = false;
+            // We should fully clear cache because we don't know
+            // what's changed until platform found Redis is unavailable
+            GlobalCacheRegion.ExpireRegion();
```
t13ka, line 71: The code like this cannot be tested or this hard to do. Potentially dangerous.
Can we use these cache manipulations with singleton service which is defined by DI container?

src/VirtoCommerce.Platform.Caching/Redis/RedisPlatformMemoryCache.cs
asvishnyakov: Already fixed. It was auto-generated this way because if non-nullable classes existence in C# 8

src/VirtoCommerce.Platform.Caching/Redis/RedisPlatformMemoryCache.cs
asvishnyakov: In platform it's always better to make methods at least protected and virtual for extensibility. You're yourself saying it has important logic. For some reason someone may be want to override this logic,

I'll add unit tests, because of this that PR is draft currently

src/VirtoCommerce.Platform.Caching/Redis/RedisPlatformMemoryCache.cs
asvishnyakov: @t13ka Fixed. I was sure I did it, so may be @tatarincev was right and I missed to upload little part of PR. Now looks good for me.

But really we should unsubscribe here just because this class can be reused someone. It's singleton, so unsubscription isn't required here strongly.

src/VirtoCommerce.Platform.Caching/Redis/RedisPlatformMemoryCache.cs
t13ka: I think methods with these modifiers will be never overridden, cuz we have ServiceCollectionExtensions with the only single implementation, and it is a quite general platform thing. If we want to generalize this, we have to think about pushing up these methods higher. In this implementation, the chance of reusing/overriding is quite small :)

src/VirtoCommerce.Platform.Caching/Redis/RedisPlatformMemoryCache.cs
asvishnyakov: It can be tested easy, because regions are cancellable.
Why it's potentially dangerous? Did you read PR description?

t13ka: Yeah, I've read.
Potentially dangerous cuz we using a self-defined singleton with static fields and methods. Better to use DI for this.
Eg: _globalCacheService.ExpireRegion();

Maybe it is out of scope, but anyway

t13ka: How are you going to test singleton with static fields? Do you want to write a special wrapper for it?

asvishnyakov: All cache regions are static. It's by design.

src/VirtoCommerce.Platform.Caching/PlatformMemoryCache.cs
asvishnyakov: Well, I returned them back. In addition to extensibility, there is also another reason: there is no other way to determine disabled cache or not. It also making code not testable.

asvishnyakov:
SonarQube code smells in new code are false-positive. I'm specially didn't made them auto-properties, because in this case we will call virtual members in constructor:

https://docs.microsoft.com/en-us/archive/blogs/ericlippert/why-do-initializers-run-in-the-opposite-order-as-constructors-part-one
https://www.jetbrains.com/help/resharper/2020.1/VirtualMemberCallInConstructor.html

@lnetrebskii @mvktsk FYI

src/VirtoCommerce.Platform.Caching/PlatformMemoryCache.cs
asvishnyakov: Fixed

yecli:
Released: https://github.com/VirtoCommerce/vc-platform/releases/tag/3.2.1

PR #1930, opened 2020-06-25
vc-build 1.1.0

krankenbro:
Parameters VersionTag and CustomTagSuffix were replaced with CustomVersionSuffix. 
If Directory.Build.Props has no VersionSuffix, then it will update with CustomVersionSuffix


lnetrebskii:
Review task https://virtocommerce.atlassian.net/browse/VP-3344 has been created

krankenbro:
Fixed an issue when props file is missing. The version was updated to 1.1.1

PR #1931, opened 2020-06-26
VP-3329: Set Start sending date for new email notifications

ilyawzrd:
### Problem
The start sending date is always today's date and time when notification was successfully sent via smtp. 

### Solution
This date should be date/time when send process started. 

### Proposed of changes
It was null all the time, so i added assignment to the corresponding property.

image

- [x] For unit tests - follow ms best practices: https://docs.microsoft.com/en-us/dotnet/core/testing/unit-testing-best-practices

### DIFF
--- a/CommonAssemblyInfo.cs
+++ b/CommonAssemblyInfo.cs
@@ -15,8 +15,8 @@ using System.Runtime.InteropServices;
 [assembly: AssemblyProduct("Virto Commerce 2.13")]
 [assembly: AssemblyCopyright("Copyright © VirtoCommerce 2011-2020")]
 
-[assembly: AssemblyFileVersion("2.13.64.0")]
-[assembly: AssemblyVersion("2.13.64.0")]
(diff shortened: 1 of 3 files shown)


CommonAssemblyInfo.cs
```diff
@@ -15,8 +15,8 @@
 [assembly: AssemblyProduct("Virto Commerce 2.13")]
 [assembly: AssemblyCopyright("Copyright © VirtoCommerce 2011-2020")]
 
-[assembly: AssemblyFileVersion("2.13.63.0")]
-[assembly: AssemblyVersion("2.13.63.0")]
+[assembly: AssemblyFileVersion("2.13.64.0")]
+[assembly: AssemblyVersion("2.13.64.0")]
```
yecli, line 19: Do not need to increase version in this commit, as it is to the dev branch.
It will be increased before merge to master - in Assembly info and in module.manifest

PR #1945, opened 2020-07-26
Fix path to storefront repository in main.yml

mvktsk:
### Problem
After renaming `storefront-core` to `storefront` CI workflow failed in swagger validation job because it uses шьфпуы шт docker.pkg.github.com/virtocommerce/vc-storefront-core/storefront 

### Solution
Fix path to storefront repository in main.yml


mvktsk:
Review task https://virtocommerce.atlassian.net/browse/VP-3799 has been created

PR #1946, opened 2020-07-27
VP-3656: development via docker

omnidark:
### Proposed of changes
VP-3656:
- Change docker-compose.override.yml to run vc-platform and mssql server in containers
- Add .env file for the customize shared folders
- Add mkdocs developer-guide


mvktsk:
Review task https://virtocommerce.atlassian.net/browse/VP-3806 has been created

Woland2k:
The entrypoint should be:
`entrypoint: ["/wait-for-it.sh", "vc-db:1433", "-t", "120", "--", "dotnet", "VirtoCommerce.Platform.Web.dll"]`

That way, platform instance waits until database is available before starting.

src/VirtoCommerce.Platform.Web/Dockerfile
```diff
@@ -1,6 +1,31 @@
...
+WORKDIR "/src/src/VirtoCommerce.Platform.Web"
+RUN dotnet build "VirtoCommerce.Platform.Web.csproj" -c Release -o /app/build
+
+FROM build AS publish
+RUN dotnet publish "VirtoCommerce.Platform.Web.csproj" -c Release -o /app/publish
+
+FROM base AS final
+WORKDIR /app
+COPY --from=publish /app/publish .
+COPY wait-for-it.sh /wait-for-it.sh
```
mvktsk, line 29: wait-for-it.sh script not available in `base` and `build` layers

mvktsk (changes requested):
1.  Remove binary `src/VirtoCommerce.Platform.Web/App_DataUploads/clothing-and-electronics_3_0_0.zip` from repository
2. `src/VirtoCommerce.Platform.Web/Dockerfile.original` looks like an unnecessary file

mvktsk (dismissed):

docker-compose.override.yml
```diff
@@ -1,12 +1,43 @@
...
+
   virtocommerce.platform.web:
-    environment:
-      - ASPNETCORE_ENVIRONMENT=Development
     ports:
-      - "80"
+      - "${DOCKER_PLATFORM_PORT:-8090}:80"
+    environment:
+     - VirtoCommerce:AllowInsecureHttp=true
+     - ConnectionStrings__VirtoCommerce=Data Source=vc-db;Initial Catalog=VirtoCommerce3docker;Persist Security Info=True;User ID=sa;Password=v!rto_Labs!;MultipleActiveResultSets=False;
```
yecli, line 24: There is SA_PASSWORD variable for the "v!rto_Labs!" password, should it be duplicated here?

.env
```diff
@@ -0,0 +1,3 @@
+DOCKER_TAG=v3
+CMS_CONTENT_VOLUME=/Users/konstantin.k/mapfolder/cms-content
```
yecli, line 2: Seems there are some personal paths here. Are they really needed in Github?

wait-for-it.sh
```diff
@@ -0,0 +1,182 @@
+#!/usr/bin/env bash
```
yecli, line 1: Maybe you could add the reference to the origin of this script

docker-compose.override.yml
KKisilevsky: This is done in order to be able to parameterize the value, as it is defined above in the service for creating mssql

wait-for-it.sh
KKisilevsky: Origin of the script https://github.com/VirtoCommerce/vc-docker/blob/master/linux/platform/wait-for-it.sh

docker-compose.override.yml
yecli: Yes, I think we should ensure password in mssql and in connection string should be the same. Otherwise changing password would require changes in 2 places. Could we make it by having one variable and using it in both places?

wait-for-it.sh
yecli: Could you add it as a comment at the beginning of the script?

yecli (approved):

mvktsk (approved):

PR #1950, opened 2020-07-30
VP-3686: Fix user phone number update

ilyawzrd:
phone


yecli:
Phone changing mechanics should be different, closing.

PR #1978, opened 2020-08-18
VP-4128: Create  workflow

krankenbro:


.github/workflows/main.yml
```diff
@@ -23,270 +23,82 @@ on:
...
+      run: vc-build Compile
 
-    - name: Install VirtoCommerce.GlobalTool
-      run: dotnet tool install --global VirtoCommerce.GlobalTool
+    - name: Unit Tests
+      run: vc-build Test -skip
 
-    - name: Download a Build Artifact
-      uses: actions/download-artifact@v2
+    - name: SonarCloud End
```
yecli, line 65: Is it ok that it is the same as 
name: SonarCloud Begin?

krankenbro: Thanks! it is typo

yecli (approved):

.github/workflows/main.yml
```diff
@@ -23,270 +23,82 @@ on:
...
+      uses: VirtoCommerce/vc-github-actions/sonar-scanner-end@dev
+
+    - name: Packaging
+      run: vc-build Compress -skip Clean+Restore+Compile+Test
+
+    #- name: Swagger Schema Validation
```
yecli, line 71: Is it OK that Swagger Schema Validation is commented out?

krankenbro: It is temporary  because of https://virtocommerce.atlassian.net/browse/VP-4255

PR #1988, opened 2020-08-24
VP-3900: Delay hangfire jobs until platform had started

akak1977:


tatarincev (commented):
I don't like this solution because it can be made by an easier way.

Under the hood, Hangfire uses the [ASP.NET HostedServices](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/host/hosted-services?view=aspnetcore-3.1&tabs=visual-studio)   to host the job processing background task. https://github.com/HangfireIO/Hangfire/blob/master/src/Hangfire.AspNetCore/HangfireServiceCollectionExtensions.cs#L147

We can consider this fact for changing the order of the running of hagfire hosted service. According to the official documentation https://docs.microsoft.com/en-us/aspnet/core/fundamentals/host/hosted-services?view=aspnetcore-3.1&tabs=visual-studio#ihostedservice-interface in order to change running hosted services after the app's pipeline has been configured need to move `AddHagfire()` in the service collection from the `Startup.cs`  to separate `ConfigureService` method is called in the `CreateHostBuilde()`  from `Program.cs`

image

PR #2006, opened 2020-09-02
VP-3682: Fix external logins

eMazeika:


src/VirtoCommerce.Platform.Web/Controllers/Api/SecurityController.cs
```diff
@@ -349,6 +349,20 @@ public async Task<ActionResult<SecurityResult>> Create([FromBody] ApplicationUse
                 result = await _userManager.CreateAsync(newUser, newUser.Password);
             }
 
+            if (result.Succeeded && !newUser.Logins.IsNullOrEmpty())
```
tatarincev, line 352: You need to move this code from here to another place with considering the two cases create and update.

For create
https://github.com/VirtoCommerce/vc-platform/blob/dev/src/VirtoCommerce.Platform.Web/Security/CustomUserManager.cs#L209

For update
https://github.com/VirtoCommerce/vc-platform/blob/dev/src/VirtoCommerce.Platform.Web/Security/CustomUserManager.cs#L177

akak1977 (approved):

src/VirtoCommerce.Platform.Web/Security/CustomUserManager.cs
```diff
@@ -174,21 +174,40 @@ public override async Task<IdentityResult> UpdateAsync(ApplicationUser user)
...
 
-                //Remove
-                foreach (var removeRole in targetRoles.Except(sourceRoles))
+                if (user.Logins != null)
                 {
-                    await RemoveFromRoleAsync(user, removeRole);
+                    var targetLogins = await GetLoginsAsync(user);
+                    var sourceLogins = user.Logins.Select(x => new UserLoginInfo(x.LoginProvider, x.ProviderKey, null));
+
+                    foreach (var item in sourceLogins.Where(x => targetLogins.All(y => x.LoginProvider + x.ProviderKey != y.LoginProvider + y.ProviderKey)))
```
tatarincev, line 202: Since the `UserLoginInfo` is derived from ValueObject you might use `.Equals()` for the more short-form of equality checking.

tatarincev (approved):

src/VirtoCommerce.Platform.Web/Security/CustomUserManager.cs
eMazeika: UserLoginInfo is for AspNetCore.Identity, not a Valueobject

PR #2042, opened 2020-09-25
VP-2262: set lowercase names for app_data, modules, certificates folders

eMazeika:


akak1977 (approved):
I've checked the clean install, import default dataset, export on both: Win and Linux.
Working ok.

tatarincev (approved):

tatarincev (commented):
I think we have to check and do such a replacement everywhere for modules projects either.
https://github.com/VirtoCommerce/vc-module-order/blob/dev/src/VirtoCommerce.OrdersModule.Web/webpack.config.js#L11

krankenbro (approved):

PR #2048, opened 2020-09-29
VP-4321: Fix add middleware for hagfire auditable tasks

Vectorfield4:
https://virtocommerce.atlassian.net/browse/VP-4321

### Proposed of changes


src/VirtoCommerce.Platform.Web/Middleware/Hangfire/HangfireUserContextMiddleware.cs
```diff
@@ -0,0 +1,53 @@
...
+using Microsoft.AspNetCore.Http;
+using VirtoCommerce.Platform.Security;
+
+namespace VirtoCommerce.Platform.Web.Middleware.Hangfire
+{
+    public class HangfireUserContextMiddleware : IClientFilter, IServerFilter
+    {
+        private readonly IHttpContextAccessor _contextAccessor;
+
+        /// <summary>
```
tatarincev, line 13: please remove this empty comment

src/VirtoCommerce.Platform.Web/Middleware/Hangfire/HangfireUserContextMiddleware.cs
```diff
@@ -0,0 +1,53 @@
+using System.Threading;
+using Hangfire.Client;
+using Hangfire.Server;
+using Microsoft.AspNetCore.Http;
+using VirtoCommerce.Platform.Security;
+
+namespace VirtoCommerce.Platform.Web.Middleware.Hangfire
+{
+    public class HangfireUserContextMiddleware : IClientFilter, IServerFilter
```
tatarincev, line 9: Add please a one-sentence description of what is main the responsibility of this class.

src/VirtoCommerce.Platform.Web/Startup.cs
```diff
@@ -477,11 +477,14 @@ public void Configure(IApplicationBuilder app, IWebHostEnvironment env)
             //Register platform settings
             app.UsePlatformSettings();
 
-            // Complete hangfire init 
+            // Complete hangfire init
             app.UseHangfire(Configuration);
 
-            app.UseModules();
+            // Add Hangfire filters/middlewares
+            var contextAccessor = app.ApplicationServices.GetRequiredService<IHttpContextAccessor>();
```
tatarincev, line 484: move these lines into `ApplicationBuilderExtensions.UseHangfire`

src/VirtoCommerce.Platform.Web/Middleware/Hangfire/HangfireUserContextMiddleware.cs
```diff
@@ -0,0 +1,81 @@
...
+
+        /// <summary>
+        /// Ctor
+        /// </summary>
+        public HangfireUserContextMiddleware(IHttpContextAccessor contextAccessor) => _contextAccessor = contextAccessor;
+
+        /// <summary>
+        /// Check if argument contains information about log operation
+        /// </summary>
+        private readonly Func<object, bool> AuditableArgument = arg => arg is OperationLog || arg is OperationLog[];
```
yecli: I think it's better to fill username user for all Hangfire tasks, not just for those that are working with OperationLog.
As we have many AuditableEntities that do not have the handler which writes to operation log.
And even for those entities I think it is better to have proper user name in CreatedBy/ModifiedBy

yecli (dismissed):

yecli (approved):

src/VirtoCommerce.Platform.Hangfire/Middleware/HangfireUserContextMiddleware.cs
```diff
@@ -0,0 +1,54 @@
...
+
+        #endregion IClientFilter Members
+
+        #region IServerFilter Members
+
+        public void OnPerforming(PerformingContext filterContext)
+        {
+            var userName = filterContext.GetJobParameter<string>(USER_NAME);
+
+            Thread.SetData(Thread.GetNamedDataSlot(USER_NAME), userName);
```
yecli, line 44: When testing the Orders change event (e.g. change Approved and save) I've not got the proper user in [PlatformOperationLog] table.
After debugging found out that it relates to the fact that thread when this middleware.OnPerforming is executed differs from the one where DbTriggers are executed https://github.com/VirtoCommerce/vc-platform/blob/5d51ae50fddded58be04a454db8daa65e2b94518/src/VirtoCommerce.Platform.Data/Extensions/ApplicationBuilderExtensions.cs#L16, so passing data using this mechanism is not working here.

yecli (changes requested):

src/VirtoCommerce.Platform.Hangfire/Middleware/HangfireUserContextMiddleware.cs
```diff
@@ -0,0 +1,62 @@
...
+        #endregion IClientFilter Members
+
+        #region IServerFilter Members
+
+        public void OnPerforming(PerformingContext filterContext)
+        {
+            var userName = filterContext.GetJobParameter<string>(USER_NAME);
+            if (userName != null)
+            {
+                _hangfireDataTransferService.UserName = userName;
```
yecli, line 51: How this solution should handle parallel hangfire task execution? 
Based on this, the same instance will be used in parallel calls - as parallelism is turned on by default in Hangfire, so Environment.ProcessorCount * 5 workers could handle jobs in parallel.
https://docs.hangfire.io/en/latest/background-processing/configuring-degree-of-parallelism.html

yecli (changes requested):

src/VirtoCommerce.Platform.Hangfire/Middleware/HangfireUserContextMiddleware.cs
yecli: I think you need to add a test on this with multiple tasks starting the jobs under different users, to see that user name persists over one job execution.

Vectorfield4:
It's not possible to achieve a solution in the case where we using a parallel task execution. I revert changes and set the filter to prevent setting up 'ModifyBy' if we do not receive a UserName from the thread data slot. This solution does not cover all situations and needs for workaround purposes.

src/VirtoCommerce.Platform.Hangfire/Middleware/HangfireUserContextMiddleware.cs
Vectorfield4: Reverted. Check last comment https://github.com/VirtoCommerce/vc-platform/pull/2048#issuecomment-702744553

src/VirtoCommerce.Platform.Security/HttpContextUserResolver.cs
```diff
@@ -15,7 +16,7 @@ public HttpContextUserResolver(IHttpContextAccessor httpContextAccessor)
 
         public string GetCurrentUserName()
         {
-            var result = Thread.CurrentPrincipal?.Identity?.Name ?? "unknown";
+            var result = Thread.GetData(Thread.GetNamedDataSlot(ThreadSlotNames.USER_NAME)) as string ?? "unknown";
```
yecli, line 19: Probably we could add a constant for "unknown" that could be used everywhere (here and in ApplicationBuilderExtensions).

yecli (approved):

PR #2049, opened 2020-09-30
VP-2517 fix sonar cube problems

Vectorfield4:
https://virtocommerce.atlassian.net/browse/VP-2517
https://virtocommerce.atlassian.net/browse/VP-2516

### Proposed of changes


src/VirtoCommerce.Platform.Core/Common/SemanticVersion.cs
```diff
@@ -137,13 +137,21 @@ public override bool Equals(object obj)
...
         {
-            int result = _version.GetHashCode();
+            var result = _version.GetHashCode();
             result = result * 31 + Prerelease.GetHashCode();
             return result;
         }
 
         #region IComparable Members
 
+        public static bool operator >=(SemanticVersion a, SemanticVersion b) => a.CompareTo(b) >= 0;
```
yecli, line 147: Seems it can be NullReference when a = null. Also is think (null, null) case should be handled.
So probably need to remove these from here, or to implement this logic.

src/VirtoCommerce.Platform.Data/ExportImport/PlatformExportImportManager.cs
```diff
@@ -147,137 +147,152 @@ private async Task ImportPlatformEntriesInternalAsync(ZipArchive zipArchive, Pla
             var batchSize = 20;
 
             var platformZipEntries = zipArchive.GetEntry(PlatformZipEntryName);
-            if (platformZipEntries != null)
+            if (platformZipEntries is null || !manifest.HandleSecurity)
```
yecli, line 147: `manifest.HandleSecurity` is not arbitrary for all cases, settings check `manifest.HandleSettings`.
I'd make a standalone ticket and PR for this ticket, as it should be tested carefully.

yecli (changes requested):

src/VirtoCommerce.Platform.Core/Common/SemanticVersion.cs
yecli: Also need to add tests for these operators, if we add them.

src/VirtoCommerce.Platform.Data/ExportImport/PlatformExportImportManager.cs
Vectorfield4: Reverted, fixed, moved to another PR https://github.com/VirtoCommerce/vc-platform/pull/2050

src/VirtoCommerce.Platform.Core/Common/SemanticVersion.cs
Vectorfield4: Added null check, added tests

yecli (approved):

akak1977 (approved):

PR #2056, opened 2020-10-05
VP-5208: Add liquid files content provider

Vectorfield4:
https://virtocommerce.atlassian.net/browse/VP-5208

### Proposed of changes

### DIFF
--- a/src/VirtoCommerce.Platform.Web/Startup.cs
+++ b/src/VirtoCommerce.Platform.Web/Startup.cs
@@ -19,6 +19,7 @@ using Microsoft.AspNetCore.Identity;
 using Microsoft.AspNetCore.Mvc;
 using Microsoft.AspNetCore.Mvc.Formatters;
 using Microsoft.AspNetCore.Server.Kestrel.Core;
+using Microsoft.AspNetCore.StaticFiles;
 using Microsoft.EntityFrameworkCore;
 using Microsoft.Extensions.Configuration;
 using Microsoft.Extensions.DependencyInjection;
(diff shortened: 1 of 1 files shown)


AliveMen (approved):

yecli (approved):

src/VirtoCommerce.Platform.Web/Startup.cs
```diff
@@ -426,7 +428,16 @@ public void Configure(IApplicationBuilder app, IWebHostEnvironment env)
 
             app.UseHttpsRedirection();
 
+            // Add default wwwroot static files bindings
             app.UseStaticFiles();
+
+            // Add additional bindings
+            var fileExtensionsBindings = Configuration.GetSection("Content").GetSection("FileExtensionsBindings").Get<Dictionary<string, string>>();
```
tatarincev, line 435: please consider of using such extension https://github.com/VirtoCommerce/vc-storefront/blob/master/VirtoCommerce.Storefront/Controllers/AssetController.cs#L92 rather than  increasing the complexity of  app settings

src/VirtoCommerce.Platform.Web/Startup.cs
```diff
@@ -426,7 +428,16 @@ public void Configure(IApplicationBuilder app, IWebHostEnvironment env)
...
+            app.UseStaticFiles(new StaticFileOptions
+            {
+                ContentTypeProvider = new FileExtensionContentTypeProvider(fileExtensionsBindings)
```
yecli, line 438: Let's add our types to existing `FileExtensionContentTypeProvider`, not to replace it.
Thus we remove all existing builtin type mappings from here:
https://github.com/dotnet/aspnetcore/blob/0e7ca749de460555da0d9c00a351f035d546ec01/src/Middleware/StaticFiles/src/FileExtensionContentTypeProvider.cs

yecli (changes requested):

yecli (approved):

PR #2057, opened 2020-10-06
VP-2037: Properly handle modules loading errors

AliveMen:
### Problem
If any module is not loaded then the platform starts with an empty screen and error in the console don't point to module with error.

### Solution
Add error handler


AliveMen:
image

Vectorfield4 (dismissed):

AliveMen:
We have only script error
image

Vectorfield4 (dismissed):

yecli (approved):

src/VirtoCommerce.Platform.Web/Views/Home/Index.cshtml
```diff
@@ -46,5 +88,63 @@
...
+    <div class="error-content" style="display:none">
+        <p style="padding-top: 10px; padding-left: 50px; font-size: 24px;">The following modules were loaded with errors:</p>
+        <div id="moduleInfo" class="error-text">
+        </div>
+    </div>
+</div>
 <ui-view></ui-view>
 
+
+<style>
```
tatarincev, line 105: can we move these styles to https://github.com/VirtoCommerce/vc-platform/tree/master/src/VirtoCommerce.Platform.Web/wwwroot/css/themes/main/sass/modules

src/VirtoCommerce.Platform.Web/Views/Home/Index.cshtml
```diff
@@ -4,20 +4,62 @@
...
+                $.ajax({
+                    type: "GET",
+                    url: "api/platform/modules/errors",
+                    success: (data) => {
+                        var modulesWithErrors = data;
+                        if (modulesWithErrors.length > 0) {
+                            $("div.error-content").show();
 
+                            modulesWithErrors.forEach(module => {
+                                $("#moduleInfo").append(` <button type="button" class="collapsible">${module.title}</button> <div class="content"> ${module.validationErrors.join(';')} </div>`);
```
tatarincev, line 39: I think there is would be better to display a full module name Id + Version

yecli (dismissed):

yecli (dismissed):

src/VirtoCommerce.Platform.Web/Controllers/Api/ModulesController.cs
```diff
@@ -84,6 +84,28 @@ public ActionResult<ModuleDescriptor[]> GetModules()
...
         }
 
+        /// <summary>
+        /// Get installed modules with errors
+        /// </summary>
+        /// <returns></returns>
+        [HttpGet]
+        [Route("errors")]
+        [AllowAnonymous]
+        public ActionResult<ModuleDescriptor[]> GetModulesErrors()
```
tatarincev, line 94: Better move this method to the `DiagnosticsController` and use `IModuleCatalog` instead of external

AliveMen: Moved

yecli (approved):

PR #2064, opened 2020-10-09
HDOT-1833 Fix disabling boolean metaform fields

avas:
### Problem
When we add any readonly boolean fields via metaforms, they do not get disabled. The reason is that the template for such fields does not mark input and the 'switch' as disabled.

### Solution
We should probably copy the [code from VC2](https://github.com/VirtoCommerce/vc-platform/blob/support/2.x-dev/VirtoCommerce.Platform.Web/Scripts/common/directives/genericValueInput.tpl.html#L82-L87) (with respect of the `isReadonly -> isReadOnly` rename).

### Proposed of changes
So, I did exactly that - now 'switches' for readonly boolean fields should be properly disabled.
image


mvktsk:
Review task https://virtocommerce.atlassian.net/browse/VP-5507 has been created

tatarincev (approved):

akak1977 (approved):

PR #2065, opened 2020-10-09
VP-0: Add doc on Swagger type inheritance

yecli:
### Problem

Add the document about exposing derived type in Swagger doc


docs/techniques/exposing-polymorphic-types-swagger.md
```diff
@@ -0,0 +1,71 @@
...
+    // ...
+}
+
+public class AnotherDerivedObject : BaseObject
+{
+    // ...
+}
+```
+
+This will expose `BaseObject`, `DerivedObject` and `AnotherDerivedObject` in Swagger API description (despite the fact that `GetObjects()` method still has only the base type in its signature), and it won't break other API. However, to use this approach, Swashbuckle configuration code needs to be changed a bit.
```
avas, line 66:
```suggestion
This will expose `BaseObject`, `DerivedObject` and `AnotherDerivedObject` in Swagger API description (despite the fact that `GetObjects()` method still has only the base type in its signature), and it won't break other API.
```
The Swashbuckle configuration does not need to be changed anymore - the required change had already been included in #2063.

docs/techniques/exposing-polymorphic-types-swagger.md
```diff
@@ -0,0 +1,71 @@
...
+On the other hand, we don't want to enable polymorphism globally, as this might break existing API (e.g. any action from `vc-module-customer` that works with a `Member` class).
+
+## Solution
+Since the creation of VC platform v3, polymorphism support in Swashbuckle [improved significantly](https://github.com/domaindrivendev/Swashbuckle.AspNetCore/pull/1792). Now we actually can use it in actions, and the resulting document will be suitable for AutoRest (e.g. to generate API clients for storefront). An example mentioned above can be reworked like this to expose derived models:
+
+```csharp
+using Swashbuckle.AspNetCore.Annotations;
+
+[SwaggerSubType(typeof(DerivedObject)]
+[SwaggerSubType(typeof(AnotherDerivedObject)]
```
avas, line 49: It might also be nice to also mention [type discriminator annotations](https://github.com/domaindrivendev/Swashbuckle.AspNetCore#enrich-polymorphic-base-classes-with-discriminator-metadata). Unfortunately, this seems to be poorly implemented in Swagger client generation tools, but this might change in the future.

avas (approved):

PR #2068, opened 2020-10-12
VP-5352: Fix vulnerability in LicensingController

Vectorfield4:
### Proposed of changes

### DIFF
--- a/src/VirtoCommerce.Platform.Web/Controllers/Api/LicensingController.cs
+++ b/src/VirtoCommerce.Platform.Web/Controllers/Api/LicensingController.cs
@@ -1,6 +1,7 @@
 using System;
 using System.IO;
 using System.Net.Http;
+using System.Text.RegularExpressions;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Authorization;
 using Microsoft.AspNetCore.Http;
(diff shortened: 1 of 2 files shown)


src/VirtoCommerce.Platform.Web/Controllers/Api/LicensingController.cs
```diff
@@ -17,29 +18,36 @@ namespace VirtoCommerce.Platform.Web.Controllers.Api
...
             {
                 var activationUrl = new Uri(_platformOptions.LicenseActivationUrl + activationCode);
                 var httpResponse = await httpClient.GetAsync(activationUrl);
-                if (httpResponse.IsSuccessStatusCode)
+
+                if (!httpResponse.IsSuccessStatusCode)
                 {
-                    var rawLicense = await httpResponse.Content.ReadAsStringAsync();
-                    license = License.Parse(rawLicense, Path.GetFullPath(_platformOptions.LicensePublicKeyPath));
+                    return BadRequest(new { Error = $"License server returns {httpResponse.StatusCode} status code" });
```
yecli, line 43: Seems that with the addition of BadRequests, client-side license should be reworked accordingly, so the user could see a proper message, not just empty error. In fact it is breaking change in the public API behavior, you need to think and check how it behaves after that.
Please check how the new code works with:
https://github.com/VirtoCommerce/vc-platform/blob/5d51ae50fddded58be04a454db8daa65e2b94518/src/VirtoCommerce.Platform.Web/wwwroot/js/app/licensing/license-detail.js#L20

yecli (changes requested):

yecli (approved):

tatarincev (approved):

PR #2069, opened 2020-10-12
VP-5353: Fix platform export import controller

Vectorfield4:
### Proposed of changes

- [ ] Check all the changes in github PR - files count (non of them are redundant, have meaningful changes, all are added), if target branch is correct
- [ ] Check methods and variable namings - it should be self descriptive, no typos
- [ ] Check you did not introduce breaking changes in API and public models/services.
- [ ] Respect extensibility - https://community.virtocommerce.com/t/extensibility-basics-the-domain-model-and-persistence-layer-extension/141
- [ ] Follow [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself) and [SOLID](https://en.wikipedia.org/wiki/SOLID) principles
- [ ] For unit tests - follow Microsoft best practices: https://docs.microsoft.com/en-us/dotnet/core/testing/unit-testing-best-practices
- [ ] Consolidate solution dependencies in case you are using newer version, update VC module dependencies in module.manifest. Do not upgrade 3rd party packages that are shipped with the platform with the version newer than in the platform.
- [ ] Check code style conventions - https://github.com/VirtoCommerce/styleguide/blob/master/csharp.md
- [ ] Check PR have a concise and descriptive title, follow [git commit message rules](https://github.com/VirtoCommerce/styleguide/blob/master/gitcommits.md)

### DIFF
--- a/src/VirtoCommerce.Platform.Web/Controllers/Api/PlatformExportImportController.cs
+++ b/src/VirtoCommerce.Platform.Web/Controllers/Api/PlatformExportImportController.cs
@@ -19,7 +19,9 @@ using VirtoCommerce.Platform.Core.PushNotifications;
 using VirtoCommerce.Platform.Core.Security;
 using VirtoCommerce.Platform.Core.Settings;
 using VirtoCommerce.Platform.Hangfire;
+
 using Permissions = VirtoCommerce.Platform.Core.PlatformConstants.Security.Permissions;
+using SystemFile = System.IO.File;
 
 namespace VirtoCommerce.Platform.Web.Controllers.Api
 {
@@ -187,7 +189,7 @@ namespace VirtoCommerce.Platform.Web.Controllers.Api
             var localPath = Path.Combine(localTmpFolder, Path.GetFileName(fileName));
 
             //Load source data only from local file system
-            using (var stream = System.IO.File.Open(localPath, FileMode.Open))
+            using (var stream = SystemFile.Open(localPath, FileMode.Open))
             {
                 var provider = new FileExtensionContentTypeProvider();
                 if (!provider.TryGetContentType(localPath, out var contentType))
@@ -369,12 +371,14 @@ namespace VirtoCommerce.Platform.Web.Controllers.Api
                 {
                     Directory.CreateDirectory(localTmpFolder);
                 }
-                if (System.IO.File.Exists(localTmpPath))
+
+                if (SystemFile.Exists(localTmpPath))
                 {
-                    System.IO.File.Delete(localTmpPath);
+                    SystemFile.Delete(localTmpPath);
                 }
+
                 //Import first to local tmp folder because Azure blob storage doesn't support some special file access mode
-                using (var stream = System.IO.File.OpenWrite(localTmpPath))
+                using (var stream = SystemFile.OpenWrite(localTmpPath))
                 {
                     var manifest = exportRequest.ToManifest();
                     await _platformExportManager.ExportAsync(stream, manifest, progressCallback, new JobCancellationTokenWrapper(cancellationToken));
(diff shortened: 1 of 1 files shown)


src/VirtoCommerce.Platform.Web/Controllers/Api/PlatformExportImportController.cs
```diff
@@ -368,12 +390,14 @@ void progressCallback(ExportImportProgressInfo x)
                 {
                     Directory.CreateDirectory(localTmpFolder);
                 }
-                if (System.IO.File.Exists(localTmpPath))
+
+                if (SystemFile.Exists(localTmpPath))
```
yecli, line 394: Seems such an alias for System.IO.File using reduce readability, it looks like some unknown new class called `SystemFile` instead of the well-known System.IO.File. So I'd better leave the previous version.

src/VirtoCommerce.Platform.Web/Controllers/Api/PlatformExportImportController.cs
```diff
@@ -208,65 +216,79 @@ public ActionResult DownloadExportFile([FromRoute] string fileName)
...
+                    if (!Uri.IsWellFormedUriString(sampleDataInfo.Url, UriKind.Absolute))
+                    {
+                        var uri = new Uri(sampleDataUrl);
+                        sampleDataInfo.Url = new Uri(uri, uri.AbsolutePath + "/" + sampleDataInfo.Url).ToString();
+                    }
                 }
+
+                retVal.AddRange(sampleDataInfos);
+
+                return retVal;
```
yecli, line 272: In case you've refactored this method, please also rename `retVal` to `result`.
Also seems return row should not be in the scope of the using - do not see the necessity in that change.

Vectorfield4: Renamed. About variable, we don't need it inside of scope, so we don't need to take out it.

src/VirtoCommerce.Platform.Web/Controllers/Api/PlatformExportImportController.cs
Vectorfield4: We cannot use the `File` class insider of the controller, because File is a method from the controller. So we have two variants:
1. Use `System.IO.File`;
2. Make an alias for it;

I think the second way is more clear.

yecli: Imho the `System.IO.File` is more clear. It is well known for almost every .NET developer, and `SystemFile` just raises the questions.
So I'd leave this as before.

Vectorfield4: Reverted.

yecli (approved):

src/VirtoCommerce.Platform.Web/Controllers/Api/PlatformExportImportController.cs
```diff
@@ -134,6 +127,21 @@ public ActionResult<PlatformExportManifest> LoadExportManifest([FromQuery] strin
                 throw new ArgumentNullException(nameof(fileUrl));
             }
 
+            // If fileUrl contains absolute path we need to prevent loading stream from this path.
```
tatarincev, line 130: I suggest simplifying these code by this simple check
```C#
Path.GetFullPath(localPath);
            //Do not allow the use paths exploration 
            //for security reason (avoid the file structure manipulation through using relative paths)
            if (!path.StartsWith(_platformOptions.LocalUploadFolderPath))
            {
                throw new PlatformException($"Invalid path {path}");
            }
```

yecli (approved):

tatarincev (approved):

PR #2071, opened 2020-10-13
HDOT-1745 Asset upload blade: allow uploading SVG images, add a way to define custom asset filters

avas:
### Problem
Current implementation of the asset upload blade has two issues:
1. If you set accepted MIME type to `image/*` or to some specific image type, it only allows to upload PNG, JPG, GIF and BMP images. You won't be able to upload SVG images, even though they are images too.
2. Even if you specify some specific type (e.g. `image/svg+xml`), the blade will allow to upload just any image with allowed type. There is no way to allow uploading specific images, e.g. SVG images only.

### Solution
1. We should add `svg+xml` to a list of allowed images;
2. Also, it would be nice to have a way to filter uploaded files by file type.

### Proposed of changes
So, I did the following changes:
1. Added `svg+xml` to a list of allowed image types;
2. Added a new optional parameter for the blade - `filterCallback`, which should be a function accepting file description from `FileUploader`. Here is an example that allows to upload only SVG images:
```js
var newBlade = {
    id: 'imageUpload',
    currentEntityId: 'svg-images',
    title: 'platform.blades.asset-upload.title',
    subtitle: 'Image upload',
    controller: 'platformWebApp.assets.assetUploadController',
    template: '$(Platform)/Scripts/app/assets/blades/asset-upload.tpl.html',
    fileUploadOptions: {
        singleFileMode: true,
        filterCallback: function(fileDescription) {
            return fileDescription.type === 'image/svg+xml';
        },
        accept: 'image/svg+xml',
        suppressParentRefresh: true
    }
};
```
However, this callback can also be used for some more advanced filtering - e.g. to verify that uploaded asset has some specific name and file size:
```js
var newBlade = {
    // ...
    fileUploadOptions: {
        filterCallback: function(fileDescription) {
            return fileDescription.name === 'some-specific-file.json' && fileDescription.size <= 512*1024;
        },
        // ...
    }
};
```

### DIFF
--- a/src/VirtoCommerce.Platform.Web/wwwroot/js/app/assets/blades/asset-upload.js
+++ b/src/VirtoCommerce.Platform.Web/wwwroot/js/app/assets/blades/asset-upload.js
@@ -18,12 +18,17 @@ angular.module('platformWebApp')
                     removeAfterUpload: true
                 });
 
-                if (blade.fileUploadOptions.accept && blade.fileUploadOptions.accept.contains('image')) {
+                if (blade.fileUploadOptions.typeFilterCallback && angular.isFunction(blade.fileUploadOptions.typeFilterCallback)) {
+                    uploader.filters.push({
+                        name: 'customFileTypeFilter',
+                        fn: blade.fileUploadOptions.typeFilterCallback
+                    });
+                } else if (blade.fileUploadOptions.accept && blade.fileUploadOptions.accept.contains('image')) {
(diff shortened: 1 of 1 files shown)


mvktsk:
Review task https://virtocommerce.atlassian.net/browse/VP-5540 has been created

akak1977 (approved):

tatarincev (approved):

PR #2078, opened 2020-10-20
Allow to reset storefront cache through the platform UI

avas:
### Problem
There was a useful command in VC2 that allowed to manually invalidate storefront cache:
image

Unfortunately, it got removed in VC3, and there is no way to manually reset the storefront cache. This makes cache issues debugging very painful, and overcoming caching issues on non-development environments becomes almost impossible.

### Solution
The platform definitely should have a command for resetting the storefront cache manually.

### Proposed of changes
So, this PR adds a 'Reset storefront cache' command to the settings blade (since the cache is not tied to any specific store - it is reset globally).
image
image

It would also be nice to have a way to reset the platform cache manually, but this might be much more difficult to develop.


mvktsk:
Review task https://virtocommerce.atlassian.net/browse/VP-5644 has been created

yecli (dismissed):

yecli (approved):

PR #2080, opened 2020-10-22
VP-3375: Add FileLock to delay platform instance startup before another instance completes copying to the probing path

akak1977:


akak1977:
This doesn't need anymore.
Implemented in https://github.com/VirtoCommerce/vc-platform/pull/2227

PR #2088, opened 2020-11-02
VP-5563: Make swagger UI look better

yecli:
Removed redundant standard swagger style overrides, left only VC-related branding stuff;


yecli:
Now:
image
Before the fix:
image

akak1977 (approved):

Vectorfield4:
if we remove reset.css we can get a different UI for different browsers. Should we remove it?

yecli:
@Vectorfield4 Returned reset.css

tatarincev (approved):

PR #2089, opened 2020-11-06
VP-5618: Remove HangFire SQL deps from AI

akak1977:


src/VirtoCommerce.Platform.Web/appsettings.json
```diff
@@ -32,6 +32,11 @@
                 "DisableGlobalLocks": true,
                 "EnableHeavyMigrations": true
             }
+        },
+        "ApplicationInsights": {
+            "IgnoreSqlTelemetryOptions": {
+                "QueryIgnoreSubstrings": [ "HangFire" ]
```
tatarincev, line 38: I think better will using a more exact pattern for  hangfire SQL queries

PR #2090, opened 2020-11-09
VP-5673: add positive integer setting value type

Vectorfield4:
### Problem:
No backend/frontend validation for setting only allows positive integers. For example import page size. At this moment we can save incorrect values to settings.

### Solution:
1. Add a positive integer value type to the backend. 
1. Save only correct values to the database, skip invalid `ObjectSettingEntry`s;
1. Add frontend validation to prevent sending incorrect values;
1. Add a positive integer template to `vaGenericValueInput` directive.

### Jira:
https://virtocommerce.atlassian.net/browse/VP-5673


tatarincev (commented):
I suggest postponing these changes. A better solution in my opinion will add validation rules for settings as it done for catalogs instead of ad-hoc solutions like this. 
image

src/VirtoCommerce.Platform.Core/Settings/SettingValueType.cs
```diff
@@ -1,14 +1,20 @@
...
+        DateTime,
+        Decimal,
+        Integer,
+        Json,
+        LongText,
+
+        /// <summary>
+        /// A000027 - Natural numbers
+        /// </summary>
+        PositiveInteger,
```
tatarincev, line 15: It is the potential breaking change, new enumeration values ​​should be defined with the highest numbers.

AliveMen (approved):

akak1977 (approved):

src/VirtoCommerce.Platform.Data/Settings/SettingsManager.cs
```diff
@@ -170,15 +172,16 @@ public virtual async Task SaveObjectSettingsAsync(IEnumerable<ObjectSettingEntry
                     .Where(x => settingNames.Contains(x.Name))
                     .ToListAsync());
 
-                foreach (var setting in objectSettings.Where(x => x.ItHasValues))
+                var validator = new ObjectSettingEntryValidator();
+                foreach (var setting in objectSettings.Where(x => x.ItHasValues).Where(x => validator.Validate(x).IsValid))
```
yecli, line 176: Here I think we should not save valid entities and skip not valid, we should throw an exception in case any value is invalid.
Otherwise, data could be saved in an unpredicted manner from the user POV.

yecli (commented):
I like the idea proposed by @tatarincev about adding validation rules as a generic solution for all validation cases. It is a good candidate for creating an epic.
A positive integer is a frequently used type of setting, so its addition (with UI validation) seems reasonable for me. It gives better UX. And it is simpler to just set this type instead of setting regexp in many cases.

yecli (approved):

PR #2111, opened 2020-12-08
VP-5618: Add entity changes collector functionality

akak1977:


src/VirtoCommerce.Platform.Core/ChangeLog/IChangeLogService.cs
```diff
@@ -1,3 +1,4 @@
+using System;
```
yecli: Is it needed?

src/VirtoCommerce.Platform.Core/ChangeLog/ILastModifiedDateTime.cs
```diff
@@ -4,7 +4,7 @@ namespace VirtoCommerce.Platform.Core.ChangeLog
 {
     public interface ILastModifiedDateTime
     {
-        DateTimeOffset LastModified { get; }
-        void Reset();
+        public DateTimeOffset GetLastModified(string entityName = null);
```
yecli, line 7: It is a breaking change in a public interface. Is it safe to do so with this one, or better to use another interface to preserve compatibility?

yecli: Also `public` is not needed in the interface declaration

akak1977: I'm sure nobody uses this call directly.

avas: No, we do use `ILastModifiedDateTime.Reset()` method to invalidate storefront cache in Heineken modules. IMO, dedicated interface for scope-based modified dates will be safer.

akak1977: I just patched this (preserved compatibility)

src/VirtoCommerce.Platform.Data/ChangeLog/ChangeLogService.cs
```diff
@@ -98,9 +76,41 @@ public virtual async Task DeleteAsync(string[] ids)
...
+
+            entityName ??= string.Empty;
+
+            var cacheKey = CacheKey.With(GetType(), "LastModifiedDateTime", entityName);
+            result = _memoryCache.GetOrCreateExclusive(cacheKey, options =>
+            {
+                options.AddExpirationToken(ChangeLogCacheRegion.CreateChangeTokenForKey(entityName));
+
+                return DateTimeOffset.UtcNow;
+            });
```
avas, line 97: If requested entity name is not used by platform, its modification date will be cached and then will never be updated. For example, if we override some entity of VC modules, storefront will also need to be updated to use the type name of new entity - otherwise, its cache will not work properly. It could be safer to always return `DateTime.UtcNow` for such cases - however, I can't think of any way to detect that.

src/VirtoCommerce.Platform.Web/Controllers/Api/ChangeLogController.cs
```diff
@@ -51,12 +52,39 @@ public ActionResult<LastModifiedResponse> GetLastModifiedDate([FromQuery] string
...
                 Scope = scope,
-                LastModifiedDate = _lastModifiedDateTime.LastModified.UtcDateTime
+                LastModifiedDate = _lastModifiedDateTime.GetLastModified().UtcDateTime
             };
 
             return Ok(result);
         }
 
+        [HttpPost]
+        [Route("~/api/changes/changedentities")]
```
avas, line 62:
```suggestion
        [Route("~/api/changes/changed-entities")]
```
IMO, `kebab-case` is easier to read for endpoint URLs.

src/VirtoCommerce.Platform.Web/Controllers/Api/ChangeLogController.cs
```diff
@@ -51,12 +52,39 @@ public ActionResult<LastModifiedResponse> GetLastModifiedDate([FromQuery] string
...
+            var result = new ChangedEntitiesResponse()
+            {
+                EntitiesResponses = changedEntitiesRequest.EntitiesNames.Select(x => new LastModifiedEntityResponse { EntityName = x, LastModifiedDate = _lastModifiedDateTime.GetLastModified(x).UtcDateTime })
+                    .Where(x => x.LastModifiedDate >= changedEntitiesRequest.After)
+            };
+            return Ok(result);
+        }
+
+        [HttpPost]
+        [Route("~/api/changes/forcechangedentities")]
```
avas, line 75:
```suggestion
        [Route("~/api/changes/force-changed-entities")]
```

src/VirtoCommerce.Platform.Core/ChangeLog/ILastModifiedDateTime.cs
yecli: Adding methods to the existing interface makes its old implementations non-compliant. Could you add another interface with new methods?

src/VirtoCommerce.Platform.Web/Controllers/Api/ChangeLogController.cs
```diff
@@ -51,12 +52,39 @@ public ActionResult<LastModifiedResponse> GetLastModifiedDate([FromQuery] string
...
 
+        [HttpPost]
+        [Route("~/api/changes/changedentities")]
+        [AllowAnonymous]
+        public ActionResult<ChangedEntitiesResponse> ChangedEntities(ChangedEntitiesRequest changedEntitiesRequest)
+        {
+            var result = new ChangedEntitiesResponse()
+            {
+                EntitiesResponses = changedEntitiesRequest.EntitiesNames.Select(x => new LastModifiedEntityResponse { EntityName = x, LastModifiedDate = _lastModifiedDateTime.GetLastModified(x).UtcDateTime })
+                    .Where(x => x.LastModifiedDate >= changedEntitiesRequest.After)
```
avas, line 69: One more thing: to get all changed entities since some date, storefront will need to pass all known entity names with each request. Would be nice to treat empty entity names list as all of known entities to reduce serialization cost and network traffic (though current implementation probably won't allow this, because there is no registry with all known entities).

src/VirtoCommerce.Platform.Data/ChangeLog/ChangeLogService.cs
```diff
@@ -98,9 +76,41 @@ public virtual async Task DeleteAsync(string[] ids)
...
                 }
                 await repository.UnitOfWork.CommitAsync();
-                ChangeLogCacheRegion.ExpireRegion();
+                Reset();
             }
         }
+        #endregion
+
+        #region ILastModifiedDateTime Members
+        public DateTimeOffset GetLastModified(string entityName = null)
```
tatarincev, line 85: Since have reset the last modified date of entities in DbTriggers, there is no sense to keep storing this code in ChangeLogService. Better to move this code to your own service.

src/VirtoCommerce.Platform.Data/Extensions/ApplicationBuilderExtensions.cs
```diff
@@ -12,8 +13,12 @@ public static class ApplicationBuilderExtensions
     {
         public static IApplicationBuilder UseDbTriggers(this IApplicationBuilder appBuilder)
         {
+            var lastChangesService = appBuilder.ApplicationServices.GetRequiredService<ILastModifiedDateTime>();
+
             Triggers<IAuditable>.Inserting += entry =>
             {
+                lastChangesService.Reset(entry.Entity.GetType().FullName);
```
tatarincev, line 20: Need to consider that DB entities can be inherited by other types. You need to use the original type name of the base type instead of the actual.

src/VirtoCommerce.Platform.Data/ChangeLog/LastChangesService.cs
```diff
@@ -0,0 +1,58 @@
...
+                return DateTimeOffset.UtcNow;
+            });
+
+            return result;
+        }
+
+        public void Reset(AuditableEntity entity)
+        {
+            var typeNames = new List<string>();
+            var entityType = entity.GetType();
```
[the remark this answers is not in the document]
avas: This method accepts `IEntity` now, so this probably won't be an issue anymore.

tests/VirtoCommerce.Platform.Tests/UnitTests/LastChangesServiceTests.cs
```diff
@@ -0,0 +1,82 @@
...
+            Assert.NotEqual(secondEntityFirstAttempt, secondEntitySecondAttempt);
+        }
+
+        [Fact]
+        public void ResetDatesForBaseEntityTypes()
+        {
+            // Arrange
+            ILastChangesService lastChangesService = new LastChangesService(GetPlatformMemoryCache());
+            var initialDateForBaseEntity = lastChangesService.GetLastModifiedDate(nameof(BaseEntity));
+            var initialDateForDerivedEntity = lastChangesService.GetLastModifiedDate(nameof(DerivedEntity));
```
artem-dudarev, line 68: Should be `typeof(...).FullName`

artem-dudarev (changes requested):

PR #2112, opened 2020-12-10
Allow to reset platform cache via endpoint call

avas:
This PR adds an endpoint which invalidates `GlobalCacheRegion` - this will purge the entire platform cache. It might be useful to reset cache after database changes by platform administrators (via direct SQL scripts) or to investigate cache issues by platform developers.

Note: I intentionally did not add this endpoint to platform UI - it should only be used when it's absolutely needed, so platform administrators MUST know why they need to reset the cache to use this endpoint. That's why IMO there should be no easy way to do it from UI.


mvktsk:
Review task https://virtocommerce.atlassian.net/browse/VP-6385 has been created

akak1977 (approved):

tatarincev (approved):

PR #2116, opened 2020-12-17
Bug/vp 6107 add object value

kostyrin:
https://virtocommerce.atlassian.net/browse/VP-6107


mvktsk:
Review task https://virtocommerce.atlassian.net/browse/VP-6437 has been created

src/VirtoCommerce.Platform.Core/ObjectValue/ObjectValueSerializedAsStringForIndexAttribute.cs
```diff
@@ -0,0 +1,6 @@
+using System;
+
+namespace VirtoCommerce.Platform.Core.ObjectValue
+{
+    public class ObjectValueSerializedAsStringForIndexAttribute : Attribute { }
```
yecli, line 5: Maybe better to `params string[] PropertyNames` parameters to allow to specify properties that need to be converted using this attribute. In case of empty param names no properties converted.

src/VirtoCommerce.Platform.Core/ObjectValue/ObjectValueJsonConverter.cs
```diff
@@ -0,0 +1,70 @@
...
+            throw new NotImplementedException();
+        }
+
+        public override void WriteJson(JsonWriter writer, object value, JsonSerializer serializer)
+        {
+            var obj = JObject.FromObject(value, new JsonSerializer() { ContractResolver = serializer.ContractResolver });
+
+            if (value != null
+                && obj.TryGetValue("value", out var objValue)
+                && obj.TryGetValue("valueType", out var objValueType))
```
yecli, line 30: Maybe we could determine value type based on the "deserialized" value, without dependency on `valueType` property

yecli (approved):

PR #2124, opened 2021-01-12
VP-6379: Feat trial registration

Vectorfield4:
### Registration dialog
image
### Configuration (if needed)
image


src/VirtoCommerce.Platform.Core/PlatformOptions.cs
```diff
@@ -41,6 +42,11 @@ public class PlatformOptions
         //server (ASOS) handles the incoming requests to arriving on non-HTTPS endpoints should be rejected or not. By default, this property is set to false to help
         //mitigate man-in-the-middle attacks.
         public bool AllowInsecureHttp { get; set; }
-    }
 
+        // Time before first registration required
+        public TimeSpan RegistrationDelay { get; set; }
```
yecli, line 47: Maybe you could add some default values for new settings here.
And it would be good to add their description into the [docs](https://github.com/VirtoCommerce/vc-platform/blob/master/docs/user-guide/configuration-settings.md)

src/VirtoCommerce.Platform.Web/appsettings.json
```diff
@@ -37,7 +37,11 @@
             "IgnoreSqlTelemetryOptions": {
                 "QueryIgnoreSubstrings": [ "[HangFire]." ]
             }
-        }
+        },
+        // Time before first registration required
+        "RegistrationDelay": "1.00:00:00",
+        // Time before registration expire
+        "RegistrationExpiration": "30.00:00:00"
```
yecli, line 44: Maybe it is better to not include these settings into appsetttings.json by default - so the user with a license would not think about what are these strange values and the user without a license cannot change them easily without looking to code.

yecli (approved):

src/VirtoCommerce.Platform.Core/PlatformOptions.cs
Vectorfield4: Good point. Updated docs.

Vectorfield4:
GTM Tag is here: https://github.com/VirtoCommerce/vc-gtm-tag-templates/blob/main/templates/registration/form.html

yecli (dismissed):

Vectorfield4:
@OlegoO 
1. template updated https://github.com/VirtoCommerce/vc-gtm-tag-templates/blob/main/templates/registration/form.html
2. PR updated.

OlegoO (approved):

PR #2128, opened 2021-01-13
#VP-6471 fix UserChangedEvent

kostyrin:
Problem: There is a reference in exists user which was updated. look at https://github.com/VirtoCommerce/vc-platform/blob/bee3173c872c57e1fd3498e99cd0318019f7afd7/src/VirtoCommerce.Platform.Web/Security/CustomUserManager.cs#L153 
and then look at 
https://github.com/VirtoCommerce/vc-platform/blob/bee3173c872c57e1fd3498e99cd0318019f7afd7/src/VirtoCommerce.Platform.Web/Security/CustomUserManager.cs#L160
existentUser is patched and the object is saved to Changed Entries as reference.
There are solutions:
- clone the existentUser and save to the Changed Entries https://github.com/VirtoCommerce/vc-platform/blob/c62fa05128d424cb2335bb79e45b444c6c0423f8/src/VirtoCommerce.Platform.Web/Security/CustomUserManager.cs#L153
- create a new instance based on existentUser (like our ToModel) and  save to the Changed Entries

Another problem is security handlers are called twice because services.AddSecurityServices called twice in Startup.
https://github.com/VirtoCommerce/vc-platform/blob/bee3173c872c57e1fd3498e99cd0318019f7afd7/src/VirtoCommerce.Platform.Web/Startup.cs#L98
https://github.com/VirtoCommerce/vc-platform/blob/bee3173c872c57e1fd3498e99cd0318019f7afd7/src/VirtoCommerce.Platform.Web/Startup.cs#L169
The solution is registration of security handlers move to `Configure(IApplicationBuilder app, IWebHostEnvironment env)` or call an extension in Configure.
https://virtocommerce.atlassian.net/browse/VP-6471


yecli (changes requested):
I thought the main purpose of this task is to have NewEnity properly updated (after save) in the `UserChangedEvent`.
Now seems `UserChangedEvent` still gets the same `ApplicationUser` model as before, which was built before the changes. So imho this was not done.
Something similar was done for orders: https://github.com/VirtoCommerce/vc-module-order/blob/dev/src/VirtoCommerce.OrdersModule.Data/Services/CustomerOrderService.cs#L181
There are fixes for other problems (better to describe them carefully in the PR title - problem, reason, fix), but looks like it is not related to VP-6471 main problem.

kostyrin:
> I thought the main purpose of this task is to have NewEnity properly updated (after save) in the `UserChangedEvent`.
> Now seems `UserChangedEvent` still gets the same `ApplicationUser` model as before, which was built before the changes (but when saving some user properties could be updated). So imho this was not done.
> Something similar was done for orders: https://github.com/VirtoCommerce/vc-module-order/blob/dev/src/VirtoCommerce.OrdersModule.Data/Services/CustomerOrderService.cs#L181
> There are fixes for other problems (better to describe them carefully in the PR title - problem, reason, fix), but looks like it is not related to VP-6471 main problem.
have added more details in the description

kostyrin:
have talked with Ivan Tselishchev and get info that need to get only difference between NewEntry and OldEntry (according to the method https://github.com/VirtoCommerce/vc-platform/blob/251759cc769d9f89e03e74d6d97b61795cd0a693/src/VirtoCommerce.Platform.Security/Handlers/LogChangesUserChangedEventHandler.cs#L62). There are no a spacial cases like in the Order Module.

yecli (approved):

yecli (approved):

OlegoO (approved):

PR #2129, opened 2021-01-19
ValueObjects with different values should not be equal and should have different cache keys

artem-dudarev:
This PR fixes the following issues that lead to incorrect equality components and duplicate cache keys:
1. `null` and `"null"` values are considered equal.
2. Empty collections are just ignored.


mvktsk:
Review task https://virtocommerce.atlassian.net/browse/VP-6687 has been created

tatarincev (approved):

tatarincev (approved):

PR #2136, opened 2021-01-26
VP-6366: Replace panel action icons

AliveMen:
Replace panel icons


eMazeika (commented):
The changes looks good. But it affects only small part of existing icons in Platform. As a result some icons will be solid, while other will stay outdated.

akak1977 (approved):

akak1977 (approved):

PR #2155, opened 2021-02-15
VP-6852: Add infinite scroll for catalog and priselist selectors

AliveMen:
### Related task:     
VP-6852 Set new pricelist assignments are only possible for the first 20 catalogs
### Problem:
Absent common component 
### Changes:
Add directive for dropdown infinite scroll

### DIFF
--- a/src/VirtoCommerce.Platform.Web/wwwroot/js/common/directives/uiScroll.tpl.html
+++ b/src/VirtoCommerce.Platform.Web/wwwroot/js/common/directives/uiScroll.tpl.html
@@ -1,0 +1,11 @@
+<div class="form-group">
+<label class="form-label">{{label | translate}}</label>
+<div class="form-input">
+    <ui-select ng-model="selectedId" required on-select="setValue($item)">
+        <ui-select-match placeholder="{{placeholder | translate}}">{{list.length > 0 ? $select.selected.name : 'Loading...'}}</ui-select-match>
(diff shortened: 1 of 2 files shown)


eMazeika (commented):
It's architectural decision if such component should be part of platform:
* is it ok to hard-code binding value to "id" and "name" as a label.
* it would be nice to see Platform actually using the directive, not just declaring
* in-depth testing is needed

src/VirtoCommerce.Platform.Web/wwwroot/js/common/directives/uiScroll.tpl.html
```diff
@@ -0,0 +1,11 @@
+<div class="form-group">
+<label class="form-label">{{label | translate}}</label>
+<div class="form-input">
+    <ui-select ng-model="selectedId" required on-select="setValue($item)">
+        <ui-select-match placeholder="{{placeholder | translate}}">{{list.length > 0 ? $select.selected.name : 'Loading...'}}</ui-select-match>
+        <ui-select-choices repeat="x.id as x in list | filter: { name: $select.search }" refresh="fetch($select)" when-scrolled="fetchNext($select)">
+            <span ng-bind-html="x.name | highlight: $select.search"></span>
```
akak1977, line 7: I think *id* is OK, but it will be good to have a way to change the display field instead of only *name*.

akak1977: Don't forget to raise the platform dependency version in modules...

Vectorfield4 (approved):

PR #2179, opened 2021-03-12
PT-326: Security accounts does not contain roles and permissions

AliveMen:
https://virtocommerce.atlassian.net/browse/PT-3


Vectorfield4 (approved):

Vectorfield4:
Expected behavior, no need to fix.

PR #2200, opened 2021-04-01
PT-642: Exclude all ignored files and *module files ...

akak1977:
There is a problem:
*vc-build Compress* command currently excludes assemblies related to only one level of modules dependencies, enlisted in *module.manifest*,
Solution:
Exclude all ignored files and *module files not related to compressing module, regardless of module dependencies.

https://virtocommerce.atlassian.net/browse/PT-28

### DIFF
--- a/build/Build.cs
+++ b/build/Build.cs
@@ -1,5 +1,6 @@
 using System;
 using System.Collections.Generic;
+using System.Text.RegularExpressions;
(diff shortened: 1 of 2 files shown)


Vectorfield4:
Picture for better understanding 😃 
image

krankenbro (approved):

PR #2219, opened 2021-04-20
PT-417: Add distributed lock to modules files probing/copying mechanism

akak1977:
Generalize/rework distributed locks.
Add distributed lock to modules files probing/copying mechanism to avoid collisions with simultaneous access to the same files in the shared storage at the platform startup time.

https://virtocommerce.atlassian.net/browse/PT-416

PR #2227, opened 2021-04-23
PT-417: Add distributed lock to modules files probing/copying

akak1977:
Generalize/rework distributed locks.
Add distributed lock to modules files probing/copying mechanism to avoid collisions with simultaneous access to the same files in the shared storage at the platform startup time.

https://virtocommerce.atlassian.net/browse/PT-416


Vectorfield4 (approved):
Already reviewed here https://github.com/VirtoCommerce/vc-platform/pull/2219.

PR #2262, opened 2021-05-14
PT-1371: Fix swagger smells

akak1977:
https://virtocommerce.atlassian.net/browse/PT-1371

### DIFF
--- a/src/VirtoCommerce.Platform.Core/Security/ApplicationUser.cs
+++ b/src/VirtoCommerce.Platform.Core/Security/ApplicationUser.cs
@@ -25,6 +25,8 @@ namespace VirtoCommerce.Platform.Core.Security
         public virtual string CreatedBy { get; set; }
         public virtual string ModifiedBy { get; set; }
         public virtual IList<Role> Roles { get; set; }
+
+        [SwaggerIgnore]
         public virtual ICollection<IdentityUserRole<string>> UserRoles { get; set; }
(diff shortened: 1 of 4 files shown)


src/VirtoCommerce.Platform.Core/Security/ApplicationUser.cs
```diff
@@ -25,6 +25,8 @@ public class ApplicationUser : IdentityUser, IEntity, IAuditable, ICloneable
         public virtual string CreatedBy { get; set; }
         public virtual string ModifiedBy { get; set; }
         public virtual IList<Role> Roles { get; set; }
+
+        [SwaggerIgnore]
```
ksavosteev, line 29: Do we need to set those attributes explicitly if we have ExcludeRedundantDepsFilter.cs which does the same?

Vectorfield4 (approved):

src/VirtoCommerce.Platform.Core/Security/ApplicationUser.cs
Vectorfield4: This filter removes a different swagger component. Maybe we can extend it.

akak1977: ExcludeRedundantDepsFilter removes model, instead of the member inside model.

PR #2267, opened 2021-05-19
PT-469: Automatic object changes gathering for specific properties

akak1977:


src/VirtoCommerce.Platform.Core/ChangesUtils/ChangesDetector.cs
```diff
@@ -0,0 +1,88 @@
...
+
+        /// <summary>
+        /// Gather changes info for objects by properties, defined in type T.
+        /// Properties to compare should be annotated with <see cref="UseInChangesDetectorAttribute"/>.
+        /// </summary>
+        /// <typeparam name="T"></typeparam>
+        /// <param name="newObj">An object</param>
+        /// <param name="oldObj">An object compare to</param>
+        /// <returns></returns>
+        public static ListDictionary<string, string> GatherByType<T>(T newObj, T oldObj)
```
Vectorfield4, line 26:
```suggestion
        public static ListDictionary<string, string> Gather<T>(T newObj, T oldObj)
```

src/VirtoCommerce.Platform.Core/ChangesUtils/ChangesDetector.cs
```diff
@@ -0,0 +1,88 @@
...
+
+        /// <summary>
+        /// Gather changes info for objects by properties, defined in type objType.
+        /// Properties to compare should be annotated with <see cref="UseInChangesDetectorAttribute"/>.
+        /// </summary>
+        /// <param name="objType">Type which properties should be used to compare</param>
+        /// <param name="newObj">An object</param>
+        /// <param name="oldObj">An object compare to</param>
+        /// <returns></returns>
+        public static ListDictionary<string, string> GatherByType(Type objType, object newObj, object oldObj)
```
Vectorfield4, line 40: I think we should make them generic too, to prevent problems in getters when `newObj` and `oldObj` have different types.

src/VirtoCommerce.Platform.Core/ChangesUtils/ChangesDetector.cs
```diff
@@ -0,0 +1,88 @@
...
+
+        /// <summary>
+        /// Gather changes info for objects by properties, defined in native type of objects.
+        /// Comparable objects should have equal type.
+        /// Properties to compare should be annotated with <see cref="UseInChangesDetectorAttribute"/>.
+        /// </summary>
+        /// <param name="newObj">An object</param>
+        /// <param name="oldObj">An object compare to</param>
+        /// <returns></returns>
+        public static ListDictionary<string, string> Gather(object newObj, object oldObj)
```
Vectorfield4, line 76: I don't think we need such a method.

tests/VirtoCommerce.Platform.Core.Tests/Common/ChangeDetectorTests.cs
```diff
@@ -0,0 +1,51 @@
+using VirtoCommerce.Platform.Core.ChangesUtils;
+using Xunit;
+using FluentAssertions;
+
+namespace VirtoCommerce.Platform.Core.Tests.Common
+{
+    public class ChangeDetectorTests
```
Vectorfield4, line 7: We need to add a test case:
1. Parent class has predefined properties with marked attributes (for ex. Prop1, Prop2);
2. Child class has own properties (Prop3, Prop4);
3. We have different values for Prop1, Prop2, Prop3, Prop4;
4. Collect changes for **Child** class;
5. Ensure we collect changes only for Prop3 and Prop4.

Vectorfield4 (changes requested):

src/VirtoCommerce.Platform.Core/ChangesUtils/ChangesDetector.cs
```diff
@@ -0,0 +1,78 @@
...
+            ); ;
+
+            foreach (var info in infos)
+            {
+                var newValue = info.Getter.Invoke(newObj, new object[0]);
+                var oldValue = info.Getter.Invoke(oldObj, new object[0]);
+
+                if (!object.Equals(newValue, oldValue))
+                {
+                    if (inherit || !inherit && !info.Inherited)
```
Vectorfield4, line 69: Condition can be merged
```suggestion
                if (!object.Equals(newValue, oldValue) && inherit || !info.Inherited)
                {
```

src/VirtoCommerce.Platform.Core/ChangesUtils/DetectChangesAttribute.cs
```diff
@@ -0,0 +1,13 @@
+using System;
+
+namespace VirtoCommerce.Platform.Core.Utils.ChangeDetector
+{
+    public sealed class DetectChangesAttribute: Attribute
```
Vectorfield4, line 5: Add info about this attribute, please.

Vectorfield4 (approved):

PR #2278, opened 2021-06-01
PT-1895: Localize countries in UI

eMazeika:
https://virtocommerce.atlassian.net/browse/PT-1895


akak1977 (approved):

src/VirtoCommerce.Platform.Web/wwwroot/js/i18n/resources/countries.js
```diff
@@ -1,273 +1,281 @@
...
+            },
+            normalize: function (id) {
+                var result = undefined;
+                if (!!id) {
+                    result = id.toUpperCase();
+                }
+                return result;
+            },
+            query: function () {
+                var results = [
```
Vectorfield4, line 20: Why this list is not deleted from the platform?

Vectorfield4 (commented):
Why this file is still in the platform `src/VirtoCommerce.Platform.Web/wwwroot/js/i18n/resources/countries.js`? For backward compatibility?

src/VirtoCommerce.Platform.Web/wwwroot/js/i18n/resources/countries.js
eMazeika: It's still the primary source of countries. Any requirement to refactor?

Vectorfield4: image
With this requirement, we don't this list in the JS file.

tatarincev: I agree with @Vectorfield4, why do we still need the hard-coded country list in this file? Why don't you move the list of countries for the English language to the default resource file, and read it as an array here instead?

tatarincev: I'm pretty sure $translate has the ability to read resource arrays for a specific path. If not, we can always read localization resources as JSON.

eMazeika: API added

src/VirtoCommerce.Platform.Web/Controllers/Api/CommonController.cs
```diff
@@ -0,0 +1,43 @@
...
+        /// </summary>
+        [HttpGet]
+        [Route("countries")]
+        [AllowAnonymous]
+        public ActionResult<List<Country>> GetCountries()
+        {
+            var cacheKey = CacheKey.With(GetType(), nameof(GetCountries));
+            var results = _memoryCache.GetOrCreateExclusive(cacheKey, (cacheEntry) =>
+            {
+                var filePath = Path.GetFullPath("app_data/countries.json");
```
Vectorfield4, line 36: Maybe we need a possibility to load countries from URL from app settings.

Vectorfield4 (approved):

src/VirtoCommerce.Platform.Web/Controllers/Api/CommonController.cs
eMazeika: it should be ok for now

PR #2283, opened 2021-06-04
PT-2067 skip compare by IDs if null in Entity Equals

ksavosteev:
https://virtocommerce.atlassian.net/browse/PT-2065

### DIFF
--- a/src/VirtoCommerce.Platform.Core/Domain/Entity.cs
+++ b/src/VirtoCommerce.Platform.Core/Domain/Entity.cs
@@ -24,9 +24,8 @@ namespace VirtoCommerce.Platform.Core.Common
             if (GetRealObjectType(this) != GetRealObjectType(obj))
                 return false;
 
-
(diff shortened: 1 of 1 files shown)


src/VirtoCommerce.Platform.Core/Domain/Entity.cs
```diff
@@ -24,9 +24,8 @@ public override bool Equals(object obj)
             if (GetRealObjectType(this) != GetRealObjectType(obj))
                 return false;
 
-
             var other = obj as Entity;
-            return other != null && Id == other.Id;
+            return other?.Id != null && Id != null && Id == other.Id;
```
Vectorfield4, line 28:
```suggestion
            return Id != null && Id == other?.Id;
```

Vectorfield4 (approved):

Vectorfield4:
Different solution.

PR #2314, opened 2021-06-25
PT-1262 Deploy DEMO Solution to Azure.

facedolls:
Added:
  Auto filling for less user input
  Services:
    Azure Search,
    Redis cache.
    Storefront.


Vectorfield4 (changes requested):
Change target branch to /dev/.

PR #2338, opened 2021-08-06
PT-2741: Refactor Default export to use Assets

eMazeika:


akak1977 (approved):

src/VirtoCommerce.Platform.Web/Controllers/Api/PlatformExportImportController.cs
```diff
@@ -188,21 +195,19 @@ public ActionResult Cancel([FromRoute] string jobId)
...
-                if (!provider.TryGetContentType(localPath, out var contentType))
-                {
-                    contentType = "application/octet-stream";
-                }
-                return PhysicalFile(localPath, contentType);
+                contentType = "application/octet-stream";
             }
+
+            using var blobStream = await _blobStorageProvider.OpenReadAsync(_blobUrlResolver.GetAbsoluteUrl(localPath));
+            var blobBytes = blobStream.ReadFully();
```
tatarincev, line 209: Why can't the stream be returned instead of being read into memory?   this can leads to high memory utilization
`File(blobStream, contentType)`

tatarincev (changes requested):
Another reason why we use local file storage to store exported data is security, because exported data may contain user-sensitive data that can be accessed via direct link if use blob storage instead of the local file system.  The problems with the multiple platform instances on kn8s clusters possible can be solved by using volumes.

PR #2347, opened 2021-08-20
PT-3746: Update api controller

eMazeika:


akak1977 (approved):

ksavosteev (approved):
Also fix namespace in ApiAuthorizationController please

src/VirtoCommerce.Platform.Web/Controllers/Api/SecurityController.cs
```diff
@@ -792,6 +800,12 @@ public Task<ActionResult<SecurityResult>> ResetCurrentUserPassword([FromBody] Re
 
         #endregion PT-788 Obsolete methods
 
+        private static void ReduceUserDetails(ApplicationUser user)
```
tatarincev, line 803: By removing `PasswordHash` and `SecurityStamp` from  ApplicationUser that returned from security endpoints you are introducing breaking changes for the all existing clients who work on the legacy Storefront version. It will lead to stop working all authentication operations such as SignIn, ResetPassords etc.

tatarincev (changes requested):

src/VirtoCommerce.Platform.Web/Controllers/Api/SecurityController.cs
tatarincev: https://github.com/VirtoCommerce/vc-storefront/blob/master/VirtoCommerce.Storefront/Domain/Security/UserStoreStub.cs#L127
https://github.com/VirtoCommerce/vc-storefront/blob/master/VirtoCommerce.Storefront/Domain/Security/UserStoreStub.cs#L242

Vectorfield4 (approved):

PR #2349, opened 2021-08-20
PT-50: Update Generic Crud SearchService

aizhuli:
## Description
Update generic crud search service by including post-read processing of search results since additional processing logic is required by implementations  in modules

### DIFF
--- a/src/VirtoCommerce.Platform.Data/GenericCrud/CrudService.cs
+++ b/src/VirtoCommerce.Platform.Data/GenericCrud/CrudService.cs
@@ -253,12 +253,11 @@ namespace VirtoCommerce.Platform.Data.GenericCrud
                         repository.Remove(entity);
                     }
                     await repository.UnitOfWork.CommitAsync();
+                    await AfterDeleteAsync(models, changedEntries);
                 }
-
+               
                 ClearCache(models);
-
-                await AfterDeleteAsync(models, changedEntries);
-
+                
                 //Raise domain events after deletion
                 await _eventPublisher.Publish(EventFactory<TChangedEvent>(changedEntries));
             }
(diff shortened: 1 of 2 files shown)


akak1977 (approved):

src/VirtoCommerce.Platform.Data/GenericCrud/CrudService.cs
```diff
@@ -257,11 +257,25 @@ public virtual async Task DeleteAsync(IEnumerable<string> ids, bool softDelete =
...
             }
         }
 
+        /// <summary>
+        /// Custom CRUD service can override to implement some actions after delete
+        /// </summary>
+        /// <param name="models"></param>
+        /// <param name="changedEntries"></param>
+        /// <returns></returns>
+        protected virtual Task AfterDeleteAsync(IEnumerable<TModel> models, IEnumerable<GenericChangedEntry<TModel>> changedEntries)
```
akak1977, line 273: Does it really need?
Why not just override _Delete_, then call _base.Delete_ inside of, then do your custom actions.

akak1977 (changes requested):

src/VirtoCommerce.Platform.Data/GenericCrud/CrudService.cs
aizhuli: If entry changed events need to be published only after all necessary actions are done, we need this method.

akak1977: Ок

akak1977 (approved):

mvktsk:
Actual PR https://github.com/VirtoCommerce/vc-platform/pull/2363

PR #2350, opened 2021-08-24
PT-3739

ksavosteev:
## Description
Screenshot_47


eMazeika (approved):

akak1977 (approved):

src/VirtoCommerce.Platform.Web/Controllers/Api/SecurityController.cs
```diff
@@ -523,13 +526,32 @@ public async Task<ActionResult> RequestPasswordReset(string loginOrEmail)
                 user = await UserManager.FindByEmailAsync(loginOrEmail);
             }
 
+            if (user == null)
```
tatarincev, line 529: Add please comment why we return 200 even if the user not found

tatarincev (changes requested):

src/VirtoCommerce.Platform.Web/Controllers/Api/SecurityController.cs
ksavosteev: added

PR #2374, opened 2021-09-16
Fix type casting while deserealization

pushnitsa:
## Description
The issue was discovered incidentally. While importing store tax providers DB threw an error with problems with a unique index. I started to research and found that some of our entities use discriminators for typecasting. And tax provider is one of that entities, so that while importing we should just call generic  Newtonsoft's deserializer method, and our Polymorphic JSON converter will cast the right type for us.

### DIFF
--- a/src/VirtoCommerce.Platform.Data/ExportImport/JsonSerializerExtensions.cs
+++ b/src/VirtoCommerce.Platform.Data/ExportImport/JsonSerializerExtensions.cs
@@ -41,10 +41,10 @@ namespace VirtoCommerce.Platform.Data.ExportImport
 
         public static async Task DeserializeJsonArrayWithPagingAsync<T>(this JsonTextReader reader, JsonSerializer serializer, int pageSize, Func<IEnumerable<T>, Task> action, Action<int> progressCallback, ICancellationToken cancellationToken)
         {
-            await reader.ReadAsync();
+            reader.Read();
             if (reader.TokenType == JsonToken.StartArray)
             {
-                await reader.ReadAsync();
+                reader.Read();
 
                 var items = new List<T>();
                 var processedCount = 0;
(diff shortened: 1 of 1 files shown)


mvktsk:
Review task https://virtocommerce.atlassian.net/browse/VP-7725 has been created

Vectorfield4 (approved):

akak1977 (approved):

src/VirtoCommerce.Platform.Data/ExportImport/JsonSerializerExtensions.cs
```diff
@@ -52,9 +52,8 @@ public static async Task DeserializeJsonArrayWithPagingAsync<T>(this JsonTextRea
                 {
                     cancellationToken.ThrowIfCancellationRequested();
 
-                    var itemType = AbstractTypeFactory<T>.TryCreateInstance().GetType();
-                    var item = serializer.Deserialize(reader, itemType);
-                    items.Add((T)item);
+                    var item = serializer.Deserialize<T>(reader);
```
tatarincev, line 55: This change breaks runtime types overriding via AbstractTypefactory since all objects will be instantiated with `T` based type and won't respect runtime overrides.

example:

Module-A
 void MyCoolMethod(A  param);

Module-B
`B` derived from `A` overrides via AbstractTypeFactory<A>.Override<A, B>()

Result (runtime)
MyCoolMethod receives `param` with type `A` instead of `B` as expected.

tatarincev: Please ignore this comment, I realized that I was mistaken

pushnitsa: You freaked me out.

PR #2430, opened 2021-12-03
Added mime enhancement for file type with same extension but different content type

ArghyadeepM:
…t header

## Description This is a enhancement of validating the image/pdf files listed as part of white list categorise to actually validate the content-type associated with it. 
Sometimes, a invalid file can be uploaded even though the extension passes the white list extensions.


mvktsk:
Review task https://virtocommerce.atlassian.net/browse/VP-7852 has been created

PR #2431, opened 2021-12-03
PT-2214: Add ability to configure login types

ksavosteev:
## Description
Added the ability to configure two main login types: with login/password and AzureAD.

ezgif-6-23ef12ac6410

Login/password is enabled by default. It can be disabled in configuration options in admin (new settings):
```JSON
    "PasswordLogin": {
        "Enabled": false
    },
```
AzureAD configuration options are provided out of the box (disabled by default):
```JSON
    "AzureAd": {
        "Enabled": true,
        "AuthenticationType": "AzureAD",
       ...
      },
```
If both login types are enabled the priority in which they are show can be controlled by the `Priority` option (lower go first):
```JSON
    "PasswordLogin": {
        "Enabled": true,
        "Priority": 0
    },
```
```JSON
    "AzureAd": {
        "Enabled": true,
        "Priority": 1
      },
```

### Jira-link:

 https://virtocommerce.atlassian.net/browse/PT-2214


src/VirtoCommerce.Platform.Web/Controllers/Api/AuthorizationController.cs
```diff
@@ -74,6 +78,15 @@ public async Task<ActionResult> Exchange()
                     });
                 }
 
+                if (!_passwordLoginOptions.Enabled && !user.IsAdministrator)
```
Vectorfield4, line 81:
```suggestion
                if (_passwordLoginOptions.Disabled && !user.IsAdministrator)
```

src/VirtoCommerce.Platform.Web/Controllers/Api/SecurityController.cs
```diff
@@ -782,6 +790,33 @@ public async Task<ActionResult<UserApiKey[]>> DeleteUserApiKeys([FromQuery] stri
...
+        /// Get allowed login types
+        /// </summary>
+        [HttpGet]
+        [Route("logintypes")]
+        [AllowAnonymous]
+        public ActionResult GetLoginTypes()
+        {
+            var options = new List<LoginType>
+            {
+                new LoginType
```
Vectorfield4, line 803: Two identical items?

src/VirtoCommerce.Platform.Web/Model/Security/PasswordLoginOptions.cs
```diff
@@ -0,0 +1,15 @@
+namespace VirtoCommerce.Platform.Web.Model.Security
+{
+    public class PasswordLoginOptions
+    {
+        /// <summary>
+        /// Determines whether the user authentication by username/password is enabled.
+        /// Always enabled unless explicitly disabled by the platform options.
+        /// </summary>
+        public bool Enabled { get; set; } = true;
```
Vectorfield4, line 9:
```suggestion
        public bool Disabled { get; set; }
```

Vectorfield4 (changes requested):
Update documentation https://github.com/VirtoCommerce/vc-platform/blob/master/docs/user-guide/configuration-settings.md

src/VirtoCommerce.Platform.Web/Controllers/Api/SecurityController.cs
ksavosteev: No, one item for each setting.

PR #2436, opened 2021-12-11
PT-4279: fix - settings is empty after saving changes

BelyakovSergey:
## Description
Store > Tax providers > Settings is empty after saving changes
### Jira-link:

https://virtocommerce.atlassian.net/browse/PT-4279


akak1977 (changes requested):
Don't merge, please!
So need to discuss these changes.

PR #2443, opened 2022-01-13
PT-5826: Obsolete hash providers replacement

getmansky:
## Description
Obsolete hash providers were replaced with static method call
### Jira-link:

https://virtocommerce.atlassian.net/browse/PT-5826


asvishnyakov (approved):
About FIPS compliance: https://github.com/dotnet/runtime/issues/46934

PR #2453, opened 2022-01-20
PT-5026: New implementation of the countries service basing on third party nuget

getmansky:
## Description
New implementation of the countries service basing on third party nuget
### Jira-link:

 https://virtocommerce.atlassian.net/browse/PT-5026

### DIFF
--- a/src/VirtoCommerce.Platform.Data/Common/FileSystemCountriesService.cs
+++ b/src/VirtoCommerce.Platform.Data/Common/FileSystemCountriesService.cs
@@ -61,7 +61,14 @@ namespace VirtoCommerce.Platform.Data.Common
 
         public Country GetByCode(string code)
         {
-            throw new NotImplementedException();
+            var country = GetCountries().FirstOrDefault(x => x.Id == code);
+
+            if (country == null)
+            {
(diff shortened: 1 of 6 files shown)


src/VirtoCommerce.Platform.Data/Common/FileSystemCountriesService.cs
```diff
@@ -53,5 +58,10 @@ public async Task<IList<CountryRegion>> GetCountryRegionsAsync(string countryId)
 
             return Array.Empty<CountryRegion>();
         }
+
+        public Country GetByCode(string code)
+        {
+            throw new NotImplementedException();
```
ksavosteev, line 64: is there a reason we can't implement this method?

getmansky: We don't use this service in any way but getting countries regions. It is possible to implement this method, but do we really need this implementation?

ksavosteev: I think we should fully implement interfaces

ksavosteev (approved):

PR #2481, opened 2022-03-16
PT-6377: Mapping users by email when register via Azure AD

getmansky:
## Description
Mapping users by email when register via Azure AD

### DIFF
--- a/src/VirtoCommerce.Platform.Web/Controllers/ExternalSignInController.cs
+++ b/src/VirtoCommerce.Platform.Web/Controllers/ExternalSignInController.cs
@@ -71,7 +71,8 @@ namespace VirtoCommerce.Platform.Web.Controllers
 
             ApplicationUser platformUser = null;
             var userName = GetUserName(externalLoginInfo);
-            var userEmail = externalLoginInfo.Principal.FindFirstValue(ClaimTypes.Email);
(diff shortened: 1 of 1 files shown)


src/VirtoCommerce.Platform.Web/Controllers/ExternalSignInController.cs
```diff
@@ -82,13 +82,15 @@ public async Task<ActionResult> SignInCallback(string returnUrl)
...
+                platformUser = await _userManager.FindByNameAsync(userName) ??
+                               await FindUserByEmail(GetUserEmail(externalLoginInfo));
+
                 if (platformUser == null)
                 {
                     platformUser = new ApplicationUser
                     {
                         UserName = userName,
-                        Email = externalLoginInfo.Principal.FindFirstValue(ClaimTypes.Email) ?? (userName.IsValidEmail() ? userName : null),
+                        Email = GetUserEmail(externalLoginInfo) ?? (userName.IsValidEmail() ? userName : null),
```
pushnitsa, line 93: What is the point to move the previous statement to the private method?

getmansky: It is made to avoid code duplication.

pushnitsa (approved):

PR #2486, opened 2022-04-04
Optimize caching in CrudService

artem-dudarev:
### Problem
When calling `GetAsync("1", "2")` and then `GetAsync("2", "1")` or `GetAsync("1")`, CrudService loads the same objects from the database multiple times and adds the same objects to the cache multiple times because it builds the cache key by joining all ids into a single string.
### Solution
Build a separate cache key for each object and add only one object per id to the cache.
### Benefits
- Reduced number of requests to the database
- Reduced memory usage


mvktsk:
Review task https://virtocommerce.atlassian.net/browse/VP-7994 has been created

OlegoO (approved):

ksavosteev (approved):

PR #2489, opened 2022-04-05
Add response group to the cache key prefix

artem-dudarev:
## Description
Add missing response group to the cache key prefix.


mvktsk:
Review task https://virtocommerce.atlassian.net/browse/VP-8000 has been created

OlegoO (approved):

PR #2544, opened 2022-08-09
PT-8520: New fields not showing up in hamburger menu

artem-dudarev:
## Description
A hamburger menu in a grid contains only those fields, that are present in the very first record in the grid.
This PR makes all fields to be serialized even if its values are `null`.

### Jira-link:
https://virtocommerce.atlassian.net/browse/PT-8520

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,4 +1,4 @@
-﻿root = true
+root = true
 
 [*]
 charset = utf-8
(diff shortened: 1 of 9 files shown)


OlegoO (approved):

tatarincev (changes requested):
This is a major change of how the platform responds to API requests. I think this problem can be solved in another way since  this solution can lead to unexpected consequences in different places. It's not worth it.

akak1977:
We decided to merge it, because of no different way to enable all the columns in lists.

PR #2556, opened 2022-08-30
VP-8164: Added UseAllOfToExtendReferenceSchemas to mark all  ref as Nullable

OlegoO:
## Description
Added Platform option UseAllOfToExtendReferenceSchemas to Extend Rest API reference schemas (using the allOf construct) so that contextual metadata. By default true.
Added Platform option IncludeOutputNullValues to Include null values when serializing Rest API objects. By default true.


akak1977 (approved):

docs/user-guide/configuration-settings.md
```diff
@@ -28,6 +28,8 @@ The configuration keys are hierarchical. This structure is most convenient to ma
...
 |  | Swagger | E.g. <br>"Swagger":<br>{<br>"Disable": true<br>}| Allows to disable swagger initialization at platform startup to prevent access to the swagger UI and documents.
+|  | UseAllOfToExtendReferenceSchemas | `true` | Extend Rest API reference schemas (using the allOf construct) so that contextual metadata.
```
artem-dudarev, line 31:
```suggestion
|  | UseAllOfToExtendReferenceSchemas | `true` | Extend Rest API reference schemas (using the allOf construct) so that contextual metadata can be applied to all parameter and property schemas.
```

src/VirtoCommerce.Platform.Core/PlatformOptions.cs
```diff
@@ -58,5 +58,15 @@ public class PlatformOptions
         /// Extensions of the files that can be uploaded to the server by the platform
         /// </summary>
         public string[] FileExtensionsWhiteList { get; set; } = new string[0];
+
+        /// <summary>
+        /// Extend Rest API reference schemas (using the allOf construct) so that contextual metadata.
```
artem-dudarev, line 63:
```suggestion
        /// Extend Rest API reference schemas (using the allOf construct) so that contextual metadata can be applied to all parameter and property schemas.
```

src/VirtoCommerce.Platform.Web/Swagger/SwaggerServiceCollectionExtensions.cs
```diff
@@ -31,7 +31,7 @@ public static class SwaggerServiceCollectionExtensions
         /// </summary>
         /// <param name="services"></param>
         /// <param name="configuration"></param>
-        public static void AddSwagger(this IServiceCollection services, IConfiguration configuration)
+        public static void AddSwagger(this IServiceCollection services, IConfiguration configuration, bool useAllOfToExtendReferenceSchemas)
```
artem-dudarev, line 34: Add `/// <param name="useAllOfToExtendReferenceSchemas"></param>`

artem-dudarev (changes requested):

artem-dudarev (approved):

PR #2640, opened 2023-04-13
Add a non-generic overload to AbstractTypeFactory.OverrideType

### LINKED ISSUE #2626: Add OverrideType version that accepts types through regular parameters.
**Problem:**

`AbstractTypeFactory.OverrideType`  uses generic parameters for both the overriding and overridden types which forces the caller to literally embed both types in the call and prevents scenarios where the old type (more likely) or both types are only referenced as `Type` objects. Workarounds include reflection or compiled lambda expressions, both an overkill.

Current signature:
https://github.com/VirtoCommerce/vc-platform/blob/287282536ea3315fc44f9775951f32bf00102587/src/VirtoCommerce.Platform.Core/Domain/AbstractTypeFactory.cs#L68

**Solution:**
Introduce and overload that either accepts old type as a `Type` parameter:
```cs
public static TypeInfo<BaseType> OverrideType<NewType>(Type oldType) where NewType : BaseType
{
    ...
}
```
or both old and new type:
```cs
public static TypeInfo<BaseType> OverrideType(Type oldType, Type newType)
{
    ...
}
```
In the latter case the `where NewType : BaseType` restriction would have to be replaced with a runtime check.

In both cases it doesn't actually make sense to pass `OldType` as generic param as it is immediately converted to a `Type` object anyway and not used in any other way whatsoever. `NewType` generic param has a little more sense as it is a part of the generic constraint, but I wouldn't consider that essential. Generally passing types as generic parameters when they are not used in method input or output is a code smell and it limits the ways the method can be called.

j-mok:
## Description

Adds an overload to `VirtoCommerce.Platform.Core.Common.AbstractTypeFactory<BaseType>.OverrideType<OldType,NewType>()` that does not use generic parameters, accepting types as regular arguments instead. This way it's much easier to override types when you don't have those types as literals at call site, but as plain `Type` objects.

Resolves #2626


src/VirtoCommerce.Platform.Core/Domain/AbstractTypeFactory.cs
```diff
@@ -62,13 +62,28 @@ public static TypeInfo<BaseType> RegisterType(Type type)
...
+        /// </summary>
+        /// <param name="oldType">The currently registered type</param>
+        /// <param name="newType">The type to override <paramref name="oldType"/> with</param>
+        /// <returns>TypeInfo instance to continue configuration through fluent syntax</returns>
+        public static TypeInfo<BaseType> OverrideType(Type oldType, Type newType)
+        {
+            if (!typeof(BaseType).IsAssignableFrom(newType))
+            {
+                throw new ArgumentException($"Only a type assignable to {typeof(BaseType)} can be used to " +
+                                            $"override {oldType}", nameof(newType));
```
artem-dudarev, line 84: Please don't split the message into two strings

artem-dudarev (changes requested):

OlegoO (approved):

PR #2643, opened 2023-04-13
Add options that control the timeout for v2->v3 db migration

j-mok:
## Description

It's possible now to control the command timeout for migrations that upgrade a V2 database to V3 (i.e. `20000000000000_UpdatePlatformV2`) using application configuration. This should work with all such migrations in all modules as long as they go through `DatabaseFacadeExtensions.MigrateIfNotApplied()`.

The proposed configuration entry is `Data:PlatformV2UpdateTimeout` and is typed as `TimeSpan`. Example configuration through environment:
```
SET Data__PlatformV2UpdateTimeout=00:05:30
```

The change is backward-compatible in both syntax and behavior, that is, no modules need to be updated and if the timeout is not configured, it will be copied over from connection timeout, as before.

It's also possible to set an infinite timeout with values 00:00:00 or -00:00:00.001 (`System.Threading.Timeout.InfiniteTimeSpan`).


j-mok:
I have updated the name of the proposed option from `PlatformV2UpdateTimeout` to `MigrationsTimeout` after realizing that the db command timeout set in `MigrateIfNotApplied()` affects - as a side effect - other migrations run with the same `DatabaseFacade`. Therefore it's better to generalize the setting as a global timeout for each command in each migration.

PR #2671, opened 2023-07-12
PT-12574: move default user creation to migration

ksavosteev:


artem-dudarev (approved):

asvishnyakov (changes requested):
Because Platform is db agnostic now, you MUST implement the same functionality for other db providers

asvishnyakov (approved):

PR #2698, opened 2023-09-28
PT-13725: add Azure App Configuration

ksavosteev:
## Description
1. Register new Azure App Configuration resource:

Screenshot_24

3. Add connection string to your AzureAppConfiguration. Add it to your appsettings.json `ConnectionStrings` section. You can find full connection string inside Settings -> Access Keys.
```JSON
  "ConnectionStrings": {
    "AzureAppConfigurationConnectionString": "Endpoint=<your Azure App endpoint>;Id=lHYD;Secret=<your key>"
  }
```

4. Register your key-values inside Operations -> Configuration Explorer.

> **_IMPORTANT:_** Existing keys values will be overwritten with ones from your Azure App Configuration.

Screenshot_25
 
6.  New options class need to be binded first inside Module.cs `Initialize` method first:

```cs
    public class Settings
    {
        public string UserName { get; set; }
    }
```

```cs
    public void Initialize(IServiceCollection serviceCollection)
    {
        serviceCollection.Configure<Settings>(Configuration.GetSection("VirtoCommerce:Settings"));
    }
```

Key values can be obtained by injecting `IOptions<YourOptionsClassName>`. In this example inject `IOptions<Settings>` for the new options, `IOptions<FrontendSecurityOptions>` for the existing ones.

7. Labels: configuration support labels based on the name of the environment. It loads all keys that have no label and keys that have label based on environments (Development, Production, etc.):

Screenshot_28

8. Auto reload: configuration supports a special key called `Sentinel` that acts as a versioning key for the list of  configuration values. The application checks the value of this key, if it changes the configuration values will be reloaded. By default the application polls it every 30 seconds.

Screenshot_29

To use current values of the configuration inject `IOptionsMonitor<T>` instead of `IOptions<T>` and access `CurrentValue` property.


OlegoO (approved):

alsconsulting:
the configuration looks good - but it is missing the options to set labels based on envionments, and also sentinel, which allows for auto updating based on changing values in app config:
eg.

 builder.AddAzureAppConfiguration(options =>
                            {
                                options
                                    .Connect(appConfigurationEndpoint)
                                    .ConfigureRefresh(refreshOptions =>
                                    {
                                        refreshOptions.Register("Sentinel", refreshAll: true);
                                    })
                                    .Select(KeyFilter.Any)
                                    .Select(KeyFilter.Any, context.HostingEnvironment.EnvironmentName);
                            });

PR #2715, opened 2023-10-24
PT-13884: Dynamic dictionary item name must be required.

OlegoO:
## Description
fix: Dynamic dictionary item name must be required. Throw InvalidCollectionItemException if One or more items in the collection have a null or empty Name. API returns BadRequest. 
fix: Fixs Name can be null or empty in DB. The update marks DynamicPropertyDictionaryItemEntity Name as required. Name with either Null or Empty value will be converted to unique id automatically.

### DIFF
--- a/src/VirtoCommerce.Platform.Web/Controllers/Api/DynamicPropertiesController.cs
+++ b/src/VirtoCommerce.Platform.Web/Controllers/Api/DynamicPropertiesController.cs
@@ -8,6 +8,7 @@ using Microsoft.AspNetCore.Mvc;
 using VirtoCommerce.Platform.Core;
 using VirtoCommerce.Platform.Core.Common;
 using VirtoCommerce.Platform.Core.DynamicProperties;
+using VirtoCommerce.Platform.Core.Exceptions;
 
 namespace VirtoCommerce.Platform.Web.Controllers.Api
 {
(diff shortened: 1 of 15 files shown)


src/VirtoCommerce.Platform.Web/Controllers/Api/DynamicPropertiesController.cs
```diff
@@ -148,7 +149,20 @@ public async Task<ActionResult<DynamicPropertyDictionaryItemSearchResult>> Searc
...
+            {
+                await _dynamicPropertyDictionaryItemsService.SaveDictionaryItemsAsync(items);
+            }
+            catch (ArgumentNullException ex)
+            {
+                return BadRequest(ex.Message);
+            }
+            catch (InvalidCollectionItemException ex)
+            {
+                BadRequest(ex.Message);
```
ksavosteev, line 162: return missed

ksavosteev (changes requested):

src/VirtoCommerce.Platform.Web/Controllers/Api/DynamicPropertiesController.cs
```diff
@@ -148,7 +149,20 @@ public async Task<ActionResult<DynamicPropertyDictionaryItemSearchResult>> Searc
...
+            }
```
artem-dudarev, line 163: There is no need to catch different types of exception if we always return BadRequest(ex.Message)

ksavosteev (approved):

artem-dudarev (approved):

PR #2768, opened 2024-03-04
Fix environment folder breaking changes for macOS in .Net8

okpm:
Fixes the following problem: After updating to .Net8 on macOS, DLL dependencies of custom modules are not loaded, because the default search folder is now wrong.

See: https://learn.microsoft.com/en-us/dotnet/core/compatibility/core-libraries/8.0/getfolderpath-unix


OlegoO (approved):

artem-dudarev (approved):

AlexTzykin:
fixed on this release
https://github.com/VirtoCommerce/vc-platform/releases/tag/3.812.0

PR #2781, opened 2024-04-16
VCST-1047: login-on-begalf for token auth

OlegoO:
## Description
feat: Adds the capability to perform login-on-behalf (impersonation) using token authentication within the XAPI framework. The new feature extends the /connect/token flow by adding support for a custom grant type called impersonate. Users can now leverage the grant_type=impersonate&user_id={user_id} parameter to set the vc_xapi_impersonated_customerid for impersonation purposes. Ensure that the account initiating the impersonation has the necessary platform:security:loginOnBehalf permission. To reset the impersonation , use the grant_type=impersonate and empty user_id.

### DIFF
--- a/src/VirtoCommerce.Platform.Web/Controllers/Api/AuthorizationController.cs
+++ b/src/VirtoCommerce.Platform.Web/Controllers/Api/AuthorizationController.cs
@@ -22,6 +22,7 @@ using VirtoCommerce.Platform.Security;
 using VirtoCommerce.Platform.Security.Authorization;
 using VirtoCommerce.Platform.Security.Model;
 using VirtoCommerce.Platform.Security.Services;
+using VirtoCommerce.Platform.Web.Extensions;
 using VirtoCommerce.Platform.Web.Model.Security;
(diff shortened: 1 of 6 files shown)


src/VirtoCommerce.Platform.Web/Controllers/Api/AuthorizationController.cs
```diff
@@ -207,6 +212,48 @@ public async Task<ActionResult> Exchange()
...
+                }
+
+                var loginOnBehalfAuthResult = await _authorizationService.AuthorizeAsync(User, null,
+                    new PermissionAuthorizationRequirement(PlatformConstants.Security.Permissions.SecurityLoginOnBehalf));
+                if (!loginOnBehalfAuthResult.Succeeded)
+                {
+                    return Forbid();
+                }
+
+                var userId = openIdConnectRequest.GetParameter("user_id")?.Value?.ToString();
```
ksavosteev, line 230: Add `user_id` param to OpenIDEndpointDescriptionFilter.cs

PR #2814, opened 2024-07-02
Fix path to nuget packages on linux when debugging

asvishnyakov:
## Description

`~` mean current user home folder on Linux, but doesn't expand as such path in ASP.NET Core. Instead, `Environment.SpecialFolder.UserProfile` should be used.

I also assume the same changes should be made for Windows, but I didn't made them "because it's already working" :)

Compare:

Windows
https://github.com/VirtoCommerce/vc-platform/blob/69a5a094b18554cd37b33e1afbc15cea4f3a350f/src/VirtoCommerce.Platform.Modules/AssemblyLoading/PlatformInformation.cs#L31
macOS
https://github.com/VirtoCommerce/vc-platform/blob/69a5a094b18554cd37b33e1afbc15cea4f3a350f/src/VirtoCommerce.Platform.Modules/AssemblyLoading/PlatformInformation.cs#L38
Linux
https://github.com/VirtoCommerce/vc-platform/blob/69a5a094b18554cd37b33e1afbc15cea4f3a350f/src/VirtoCommerce.Platform.Modules/AssemblyLoading/PlatformInformation.cs#L45

### Artifact URL:

Image tag:
3.843.0-pr-2814-2078-nuget-packages-on-linux-2078a5f4


basilkot (approved):

ekuvirto:
https://github.com/VirtoCommerce/vc-platform/releases/tag/3.843.0

PR #2901, opened 2025-03-14
VST-2898: Update license activation logic and add logging

OlegoO:
## Description
fix: Enhanced error handling with logging for exceptions during activation.

### Jira-link:

https://virtocommerce.atlassian.net/browse/VCST-2898
### Artifact URL:

Image tag:
ghcr.io/VirtoCommerce/platform:3.882.0-pr-2901-305f-vcst-2898-305fb3c1

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -9,7 +9,7 @@
   </PropertyGroup>
 
   <PropertyGroup>
-    <VersionPrefix>3.881.0</VersionPrefix>
+    <VersionPrefix>3.882.0</VersionPrefix>
     <VersionSuffix></VersionSuffix>
     <VersionSuffix Condition=" '$(VersionSuffix)' != '' AND '$(BuildNumber)' != '' ">$(VersionSuffix)-$(BuildNumber)</VersionSuffix>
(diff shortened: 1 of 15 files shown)


src/VirtoCommerce.Platform.Web/Controllers/Api/LicensingController.cs
```diff
@@ -21,40 +22,65 @@ public class LicensingController : Controller
...
 
             License license = null;
 
-            using (var httpClient = new HttpClient())
+            var activationUrl = new Uri(_platformOptions.LicenseActivationUrl + activationCode);
+            try
             {
-                var activationUrl = new Uri(_platformOptions.LicenseActivationUrl + activationCode);
-                var httpResponse = await httpClient.GetAsync(activationUrl);
+                using (var httpClient = new HttpClient())
```
Ljutyj, line 55: If we make a refactoring It'll be better not to use the constructor for HttpClient. Instead of it .ner core has IHttpClientFactory or opportunity to inject IHttpClient to the constructor (needs also to declare the class via AddHttpClient in IColletionServices ) for better performance.

artem-dudarev (approved):

Issue #3027, opened 2026-05-08
GetAvailableShippingRates returns 500 (ArgumentNullException) for transiently-null cart reads under multi-instance load


hgabreu:
## Summary

`CartModuleController.GetAvailableShippingRates(cartId)` (and its siblings using `ShoppingCartBuilder.TakeCart`) throw `ArgumentNullException` and surface as HTTP 500 when `IShoppingCartService.GetByIdAsync(cartId)` legitimately or transiently returns null. Two cooperating defects:

1. **`vc-module-cart`** — `GetAvailableShippingRates` / `GetAvailablePaymentMethods` / `GetAvailableShippingRatesByContext` have no null-guard on the cart they pass to `TakeCart`. `ShoppingCartBuilder.TakeCart(null)` throws `ArgumentNullException` and the unhandled exception becomes a raw 500.
2. **`vc-platform`** — `MemoryCacheExtensions.GetOrLoadByIdsAsync` (used by `CrudService<>.GetAsync`) caches null aggregates by default (`GetOrCreateExclusive`'s `cacheNullValue` parameter defaults to `true`). A single transient empty DB read for a cart Id poisons that instance's cache for up to the entry's TTL. In a multi-instance deployment without (or before) a working Redis backplane propagating `ExpireTokenForKey` cancellations, every subsequent read on the poisoned instance returns null, and every such read 500s.

The combination produces a per-cart "stuck failure" pattern: under concurrent load some carts return 500 from `availshippingrates` for the lifetime of the cached null, even though the cart row exists and is `IsDeleted=false` in the database.

## Affected versions

- `vc-module-cart` 3.840.0 — confirmed.
- Current `dev` branch of both `vc-module-cart` and `vc-platform` — confirmed by inspection (no commit since 3.840.0 changes the controller, the builder, or the cache extension method along these paths).

Earlier 3.x releases are likely also affected; we have not bisected.

## Reproduction

Deterministic, no concurrent load required. Tested in a local 3-instance platform stack (each instance with its own `IPlatformMemoryCache`, no Redis backplane) round-robined behind a load balancer. Three instances are labelled A, B, C below.

1. Create a cart `C` via instance A.
2. Read `GET /api/carts/C/availshippingrates` from each of A, B, C — all return 200 with rates. (Cart is cached on each instance.)
[...]

PR #3029, opened 2026-05-08
fix: don't cache null aggregates in GetOrLoadByIdsAsync

hgabreu:
## Description

`MemoryCacheExtensions.GetOrLoadByIdsAsync<T>` caches a null entry whenever `loadItems` returns no row for an id, because `GetOrCreateExclusive`'s `cacheNullValue` parameter defaults to `true`. On a multi-instance deployment without (or before) a working `RedisPlatformMemoryCache` backplane, a single transient empty DB read poisons that instance's local cache for up to the entry's TTL — every subsequent read for the same id returns the cached null until either every instance handles a write to it, the entry expires, or the backplane catches up. A cooperating defect in `vc-module-cart` then turns each cached null into a 500 on `availshippingrates` / `availpaymentmethods`. Full repro and analysis in #3027.

This change passes `cacheNullValue: false` at the call site so by-id lookups never cache "not found". By-id endpoints aren't at significant risk of high-volume probes for non-existent ids — the DB cost of a miss is bounded — and removing the cached null eliminates the stuck-failure pattern entirely, regardless of backplane health or write-path coverage.

The companion controller-layer fix is in VirtoCommerce/vc-module-cart#187.

### Tests

Added two tests in `MemoryCacheExtensionTests`:

- `GetOrLoadByIdsAsync_DoesNotCacheMissingIds` — would fail without this change; asserts `loadItems` is invoked again on a re-read of an id that resolved to nothing.
- `GetOrLoadByIdsAsync_CachesPresentIds` — regression guard for the present-id path; asserts a single `loadItems` invocation across reads.

### DIFF
--- a/src/VirtoCommerce.Platform.Core/Caching/MemoryCacheExtensions.cs
+++ b/src/VirtoCommerce.Platform.Core/Caching/MemoryCacheExtensions.cs
@@ -47,12 +47,15 @@ namespace VirtoCommerce.Platform.Core.Caching
                         {
                             var cacheKey = CacheKey.With(keyPrefix, id);
 
+                            // Do not cache "not found" — a transient empty read would otherwise poison
+                            // this instance's cache for the entry's TTL, manifesting as a per-id stuck
+                            // failure when no cross-instance invalidation backplane is in place.
                             result[id] = memoryCache.GetOrCreateExclusive(cacheKey, options =>
                             {
                                 var item = itemsByIds.GetValueSafe(id);
                                 configureCache(options, id, item);
                                 return item;
-                            });
+                            }, cacheNullValue: false);
                         }
                     }
                 }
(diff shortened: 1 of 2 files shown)


OlegoO:
Thanks for the detailed analysis @hgabreu — the repro and the cooperating cart defect are well documented.
Our concern is that this patches one symptom at the call site rather than addressing the root cause. The stuck-null pattern you describe is fundamentally a cache invalidation problem on multi-instance deployments, and Virto Commerce already provides the building blocks to solve it systemically:

RedisPlatformMemoryCache backplane — propagates ExpireTokenForKey cancellations across instances, which is the supported mechanism for keeping local caches coherent in a scaled-out topology. A transient empty read poisoning an instance for the full TTL is the exact failure this is meant to prevent.
Cache region / scope invalidation (CacheKey, GenericCachingRegion, ExpireRegion) — gives explicit control over what gets evicted on a write, rather than relying on per-call cacheNullValue toggles spread across call sites.

We'd recommend treating the missing/unhealthy backplane as the defect to fix, and validating the scenario in [#3027](https://github.com/VirtoCommerce/vc-platform/issues/3027) with a working Redis backplane configured.

PR #3069, opened 2026-06-29
VCST-5361: optimize ValueObject equality/hashing and ReflectionUtility hot paths

alexeyshibanov:
## Problem

`ValueObject.GetEqualityComponents()` extracts every property value via `PropertyInfo.GetValue` reflection on **every** `Equals`, `GetHashCode`, and `GetCacheKey`. When `ValueObject`-derived types are used at scale in `HashSet`, `Distinct()`, or as dictionary keys (e.g. search indexing over documents × fields × settings), the per-call reflection plus the `IEnumerable<object>` boxing/iterator overhead saturates CPU and allocates heavily.

The same shape affects `ReflectionUtility` — most importantly `GetFlatObjectsListWithInterface`, used on catalog/order/cart save/load paths to flatten entity graphs — which reflects over `GetProperties()` for every visited node on every call.

### ValueObject

Adds a per-type fast path for `Equals`/`GetHashCode` built from compiled per-property delegates:

- **Value-typed** properties use `HashCode.Add<T>` and `EqualityComparer<T>.Default.Equals` — no boxing.
- **Reference/list** properties fetch the value as `object` (free for reference types) and run the same runtime list / scalar / null logic as before (the `'['`/`']'`-marked component stream is preserved for list-bearing values).
- The fast path is used **only** for types that override neither `GetEqualityComponents()` nor `GetProperties()`. Types that customize either keep the existing component-based path unchanged.
- `GetEqualityComponents()` (used by `GetCacheKey`/`ToString` and the legacy path) now uses a cached compiled getter instead of `PropertyInfo.GetValue`.

`Equals` results are unchanged. `GetHashCode` now uses `System.HashCode` for better distribution; hash **values** change but the contract holds (equal objects produce equal hashes, stable within a run). Hash codes are not persisted — `GetCacheKey` is the stable string key and is unchanged.

#### Note on value-typed equality

Value-typed properties compare via `EqualityComparer<T>.Default` rather than boxing to `object.Equals`. For all common value types (numbers, `bool`, `DateTime`, `Guid`, enums) these are equivalent. A struct whose `IEquatable<T>.Equals` deliberately disagrees with its `Equals(object)` override would compare via the former — the idiomatic choice for a no-boxing path.

### ReflectionUtility

Caches the allocating *metadata* reflection these helpers repeat on every call (`GetInterfaces()`, `GetProperties()`, `GetCustomAttributes()` each allocate an array); the cheap pointer-walks such as `BaseType` are left uncached.

- `GetFlatObjectsListWithInterface` builds a per-`(source type, target type)` **flatten plan** once — which properties to recurse into vs. enumerate — instead of reflecting over `GetProperties()` for every node on every call. Traversal order and dedup behaviour are unchanged.
- Per-type caches added for `GetTypeInheritanceChain(To)`, `FindPropertiesWithAttribute`, `IsDictionary`, and `IsAssignableFromGenericList` (the last is also on the `ValueObject` list-component path).
- `IsDerivativeOf` now delegates to `Type.IsSubclassOf` — equivalent to the old base-class walk but allocation-free, so it is intentionally **not** cached. This is a strict base-class check: a type is no longer treated as a derivative of itself (matching `IsSubclassOf`).

## Benchmarks

`benchmarks/VirtoCommerce.Platform.Benchmark` (consolidates the former `Benchmark.AbstractTypeFactory` and `Benchmark.ArrayVsList` projects, added to the solution). Default job, .NET 10, baseline `3.1042.0` → branch.

**`ValueObjectBenchmarks`, N = 10000:**

| Operation | Time (before → after) | Allocated (before → after) |
|-----------|-----------------------|----------------------------|
| `GetHashCode` (sum over list) | 3,297 μs → 403 μs (**−88%**) | 9,641 KB → 195 KB (**−98%**) |
| `Distinct().Count()` | 3,749 μs → 579 μs (**−85%**) | 9,839 KB → 393 KB (**−96%**) |
| `Equals` (value-equal pairs) | 6,822 μs → 905 μs (**−87%**) | 19,282 KB → 1,406 KB (**−93%**) |
| `GetCacheKey` (total length) | 5,330 μs → 3,268 μs (**−39%**) | 14,762 KB → 8,908 KB (**−40%**) |

**`GetFlatObjectsBenchmarks` (entity-graph flatten), 500 nodes:**

| Operation | Time (before → after) | Allocated (before → after) |
|-----------|-----------------------|----------------------------|
| `GetFlatObjectsListWithInterface<IEntity>` | 87.4 μs → 38.4 μs (**−56%**) | 131.7 KB → 69.2 KB (**−47%**) |

`GetCacheKey` improves less than the others because it still builds a `|`-delimited string per component — its dominant cost is string allocation, not reflection, and the no-boxing fast path does not apply to it.

## Tests

- Extended `ValueObjectUnitTests` with value-typed / nullable / hash-stability / mutation / `GetProperties`-override coverage.
- Added `ReflectionUtilityTests` covering the flatten traversal contract (self / nested / array / collection-member / cycle dedup), `IsAssignableFromGenericList` / `IsDictionary` theories, inheritance chains, `FindPropertiesWithAttribute`, and the `IsDerivativeOf` base-class semantics.

All existing `ValueObject`, Core, and Platform tests pass.

---

> [!NOTE]
> **Medium Risk**
> Touches widely used core primitives (`ValueObject`, entity-graph flattening) with intentional `GetHashCode` value changes and a small `IsDerivativeOf` semantic shift; mitigated by extensive tests and preserved equality/`GetCacheKey` contracts.
> 
> **Overview**
> Replaces per-call reflection on **`ValueObject`** `Equals`/`GetHashCode` with a per-type fast path built from compiled property accessors (value types via `EqualityComparer<T>`/`HashCode.Add<T>`; lists keep the existing `'['`/`']'` component contract). Types that override `GetEqualityComponents` or `GetProperties` still use the legacy path; `GetCacheKey`/`ToString` use cached compiled getters instead of `PropertyInfo.GetValue`. **`GetHashCode` numeric values change** (now `System.HashCode`); equality and `GetCacheKey` semantics are preserved.
> 
> **`ReflectionUtility`** caches flatten plans for `GetFlatObjectsListWithInterface` (no `GetProperties` per node), plus caches for inheritance chains, attributed properties, `IsDictionary`, and `IsAssignableFromGenericList`. **`IsDerivativeOf`** now uses `Type.IsSubclassOf` (a type is no longer considered a derivative of itself).
> 
> Benchmarks move into **`benchmarks/VirtoCommerce.Platform.Benchmark`** (replacing separate test-folder benchmark projects); unit tests expand coverage for flatten traversal, caches, and value-object edge cases.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 632c53cbfdb370d464b9e3932ba765659067792e. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

Image tag:
ghcr.io/VirtoCommerce/platform:3.1043.0-pr-3069-632c-vcst-5361-valueobject-equality-reflection-632c53cb
Jira-link:
https://virtocommerce.atlassian.net/browse/VCST-5361


benchmarks/VirtoCommerce.Platform.Benchmark/ArrayVsList/ArrayVsListBenchmarks.cs
```diff
@@ -0,0 +1,47 @@
...
+    public List<string> BenchArrayToList() => DataArray.ToList();
+
+    [Benchmark]
+    public string[] BenchArrayToArray() => DataArray.ToArray();
+
+    [Benchmark]
+    public List<string> BenchListToList() => DataList.ToList();
+
+    [Benchmark]
+    public string[] BenchListoArray() => DataList.ToArray();
```
DmitryGrishinVirtoworks:
```suggestion
    public string[] BenchListToArray() => DataList.ToArray();
```

benchmarks/VirtoCommerce.Platform.Benchmark/ArrayVsList/InMemCtx.cs
```diff
@@ -0,0 +1,35 @@
...
+    public InMemCtx(DbContextOptions options) : base(options)
+    {
+    }
+
+    protected override void OnModelCreating(ModelBuilder modelBuilder)
+    {
+        modelBuilder.Entity<Something>(o =>
+       {
+           o.Property(o => o.Id);
+       }
```
DmitryGrishinVirtoworks:
```suggestion
       });
```
and remove line below

src/VirtoCommerce.Platform.Core/Common/ValueObject.cs
```diff
@@ -2,105 +2,357 @@
...
+    public static bool operator !=(ValueObject left, ValueObject right)
+    {
+        return !Equals(left, right);
+    }
+
+    public override string ToString()
+    {
+        var properties = GetProperties().Select(x => $"{x.Name}: {GetPropertyGetter(x)(this)}");
+
+        return $"{string.Join(", ", properties)}";
```
DmitryGrishinVirtoworks: is it expected that before we have:
return $"{{{string.Join(", ", GetProperties().Select(f => $"{f.Name}: {f.GetValue(this)}"))}}}";

and now:
return $"{string.Join(", ", properties)}";

as I understand the format is different now
"Name: foo, Value: bar" instead of old "{Name: foo, Value: bar}"

benchmarks/VirtoCommerce.Platform.Benchmark/ReflectionUtility/GetFlatObjectsBenchmarks.cs
```diff
@@ -0,0 +1,69 @@
...
+{
+    [Params(50, 500)]
+    public int NodeCount { get; set; }
+
+    private BenchEntity _root;
+
+    [GlobalSetup]
+    public void Setup()
+    {
+        _root = new BenchEntity { Id = "root", Name = "r", Code = "c", Tags = new List<string> { "a", "b" }, Children = new List<BenchEntity>() };
```
DmitryGrishinVirtoworks:
```suggestion
    _root = new BenchEntity { 
        Id = "root", 
        Name = "r", 
        Code = "c", 
        Tags = new List<string> { "a", "b" }, 
        Children = new List<BenchEntity>() 
    };
```

src/VirtoCommerce.Platform.Core/Common/ReflectionUtility.cs
```diff
@@ -1,220 +1,265 @@
...
+    private static readonly ConcurrentDictionary<Type, bool> _isDictionaryCache = new();
+    private static readonly ConcurrentDictionary<Type, Type[]> _typeInheritanceChains = new();
+    private static readonly ConcurrentDictionary<(Type Type, Type ToBaseType), Type[]> _typeInheritanceChainsTo = new();
+    private static readonly ConcurrentDictionary<(Type Type, Type Attribute, bool Inherit), PropertyInfo[]> _propertiesWithAttribute = new();
 
-        public static string GetPropertyName<T>(Expression<Func<T, object>> propertyExpression)
+    public static IEnumerable<string> GetPropertyNames<T>(params Expression<Func<T, object>>[] propertyExpressions)
+    {
+        var retVal = new List<string>();
+        foreach (var propertyExpression in propertyExpressions)
```
DmitryGrishinVirtoworks, line 26:
```suggestion

        foreach (var propertyExpression in propertyExpressions)
```

src/VirtoCommerce.Platform.Core/Common/ReflectionUtility.cs
```diff
@@ -1,220 +1,265 @@
...
-        /// Search all type properties that have specific attribute
-        /// </summary>
-        /// <param name="type">The type which properties need to be sought</param>
-        /// <param name="attribute">The attribute to search</param>
-        /// <param name="inherit">Should search or not thru type ancestors properties</param>
-        /// <returns></returns>
-        public static PropertyInfo[] FindPropertiesWithAttribute(this Type type, Type attribute, bool inherit)
+        LambdaExpression lambda = propertyExpression;
+        MemberExpression memberExpression;
+        if (lambda.Body is UnaryExpression unaryExpression)
```
DmitryGrishinVirtoworks, line 43:
```suggestion

        if (lambda.Body is UnaryExpression unaryExpression)
```

src/VirtoCommerce.Platform.Core/Common/ReflectionUtility.cs
```diff
@@ -1,220 +1,265 @@
...
+            memberExpression = (MemberExpression)unaryExpression.Operand;
         }
-
-        public static bool IsHaveAttribute(this PropertyInfo propertyInfo, Type attribute)
+        else
         {
-            return propertyInfo.GetCustomAttributes(attribute, true).Any();
+            memberExpression = (MemberExpression)lambda.Body;
         }
+        var retVal = memberExpression.Member.Name;
```
DmitryGrishinVirtoworks, line 51:
```suggestion

        var retVal = memberExpression.Member.Name;
```

src/VirtoCommerce.Platform.Core/Common/ReflectionUtility.cs
```diff
@@ -1,220 +1,265 @@
...
+        return _typeInheritanceChains.GetOrAdd(type, static inputType =>
         {
-            var retVal = new List<Type>();
+            var retVal = new List<Type> { inputType };
 
-            retVal.Add(type);
-            var baseType = type.BaseType;
-            while (baseType != typeof(Entity) && baseType != typeof(object))
+            var baseType = inputType.BaseType;
+            while (baseType is not null && baseType != typeof(Entity) && baseType != typeof(object))
```
DmitryGrishinVirtoworks, line 96:
```suggestion

            while (baseType is not null && baseType != typeof(Entity) && baseType != typeof(object))
```

src/VirtoCommerce.Platform.Core/Common/ReflectionUtility.cs
```diff
@@ -1,220 +1,265 @@
...
+        return _typeInheritanceChainsTo.GetOrAdd((type, toBaseType), static key =>
         {
-            var retVal = new List<Type>();
+            var retVal = new List<Type> { key.Type };
 
-            retVal.Add(type);
-            var baseType = type.BaseType;
-            while (baseType != toBaseType && baseType != typeof(object))
+            var baseType = key.Type.BaseType;
+            while (baseType is not null && baseType != key.ToBaseType && baseType != typeof(object))
```
DmitryGrishinVirtoworks, line 113:
```suggestion

            while (baseType is not null && baseType != key.ToBaseType && baseType != typeof(object))
```

src/VirtoCommerce.Platform.Core/Common/ReflectionUtility.cs
```diff
@@ -1,220 +1,265 @@
...
-        public static bool IsDerivativeOf(this Type type, Type typeToCompare)
+    public static bool IsDictionary(this Type type)
+    {
+        ArgumentNullException.ThrowIfNull(type);
+
+        return _isDictionaryCache.GetOrAdd(type, static inputType =>
         {
-            if (type == null)
+            var retVal = typeof(IDictionary).IsAssignableFrom(inputType);
+            if (!retVal && inputType is not null)
```
DmitryGrishinVirtoworks:
```suggestion

            if (!retVal && inputType is not null)
```

src/VirtoCommerce.Platform.Core/Common/ReflectionUtility.cs
```diff
@@ -1,220 +1,265 @@
...
+        });
+    }
+
+    public static bool IsAssignableFromGenericList(this Type type)
+    {
+        // Cached per type: GetInterfaces() allocates a Type[] on every call, and this runs on the
+        // ValueObject equality/hash hot path (once per list-typed component).
+        return _assignableFromGenericListCache.GetOrAdd(type, static inputType =>
+        {
+            foreach (var intType in inputType.GetInterfaces())
```
DmitryGrishinVirtoworks, line 153: Wouldn't it better if we replace it with LINQ?

src/VirtoCommerce.Platform.Core/Common/ValueObject.cs
```diff
@@ -2,105 +2,357 @@
...
+
+        if (GetType() != obj.GetType())
+        {
+            return false;
+        }
+
+        var accessors = GetComponentAccessors(GetType());
+        if (accessors is not null)
+        {
+            foreach (var accessor in accessors)
```
DmitryGrishinVirtoworks, line 55: Please take a look SonarQube issue

src/VirtoCommerce.Platform.Core/Common/ValueObject.cs
```diff
@@ -2,105 +2,357 @@
...
+    }
+
+    public override int GetHashCode()
+    {
+        var accessors = GetComponentAccessors(GetType());
+        if (accessors is not null)
         {
-            unchecked
+            var hash = new HashCode();
+            foreach (var accessor in accessors)
```
DmitryGrishinVirtoworks, line 77:
```suggestion

            foreach (var accessor in accessors)
```

src/VirtoCommerce.Platform.Core/Common/ValueObject.cs
```diff
@@ -2,105 +2,357 @@
...
+    // yields but WITHOUT its enumerator allocation — this runs on the GetHashCode hot path, so the
+    // expansion is inlined here on purpose. Keep the '['/']' list contract in lock-step with
+    // ExpandComponent and ReferenceComponentsEqual: a list expands to '[' + children + ']', and any
+    // scalar/null is added directly.
+    private static void HashReferenceComponent(object value, ref HashCode hash)
+    {
+        if (IsListComponent(value))
+        {
+            hash.Add('[');
+            foreach (var child in (IEnumerable)value)
```
DmitryGrishinVirtoworks, line 213:
```suggestion

            foreach (var child in (IEnumerable)value)
```

src/VirtoCommerce.Platform.Core/Common/ValueObject.cs
```diff
@@ -2,105 +2,357 @@
...
+    {
+        return value is not null && value.GetType().IsAssignableFromGenericList();
+    }
+
+    private static IEnumerable<object> ExpandComponent(object value)
+    {
+        if (IsListComponent(value))
+        {
+            yield return '[';
+            foreach (var child in (IEnumerable)value)
```
DmitryGrishinVirtoworks, line 255:
```suggestion

            foreach (var child in (IEnumerable)value)
```

tests/VirtoCommerce.Platform.Core.Tests/Common/ReflectionUtilityTests.cs
```diff
@@ -38,5 +70,111 @@ public void GetFlatObjectsListWithInterface_Array()
...
         }
+
+        [Fact]
+        public void GetFlatObjectsListWithInterface_IgnoresValueTypedAndStringProperties()
+        {
+            // Number/Flag/Name must not break traversal (they are skipped, not read+boxed), and the
+            // nested entity is still found via the single-object branch.
+            var node = new Node { Name = "x", Number = 42, Flag = true, Single = new Node { Id = "child" } };
+            var result = node.GetFlatObjectsListWithInterface<IEntity>();
+            Assert.Equal(2, result.Length);
```
DmitryGrishinVirtoworks, line 80:
```suggestion

            Assert.Equal(2, result.Length);
```

tests/VirtoCommerce.Platform.Core.Tests/Common/ReflectionUtilityTests.cs
```diff
@@ -38,5 +70,111 @@ public void GetFlatObjectsListWithInterface_Array()
...
+        }
+
+        [Fact]
+        public void GetFlatObjectsListWithInterface_CollectionElementsNotMatchingT_AreIgnored()
+        {
+            var container = new { Items = new[] { "a", "b" } };
+            var result = container.GetFlatObjectsListWithInterface<IEntity>();
+            Assert.Empty(result);
```
DmitryGrishinVirtoworks, line 89:
```suggestion

            Assert.Empty(result);
```

tests/VirtoCommerce.Platform.Core.Tests/Common/ReflectionUtilityTests.cs
```diff
@@ -38,5 +70,111 @@ public void GetFlatObjectsListWithInterface_Array()
...
+        }
+
+        [Fact]
+        public void GetTypeInheritanceChainTo_StopsAtTargetBaseType()
+        {
+            var toEntity = typeof(LeafNode).GetTypeInheritanceChainTo(typeof(Entity));
+            Assert.Equal(new[] { typeof(LeafNode), typeof(MidNode) }, toEntity);
+
+            var toObject = typeof(LeafNode).GetTypeInheritanceChainTo(typeof(object));
+            Assert.Equal(new[] { typeof(LeafNode), typeof(MidNode), typeof(Entity) }, toObject);
```
DmitryGrishinVirtoworks:
```suggestion

            Assert.Equal(new[] { typeof(LeafNode), typeof(MidNode), typeof(Entity) }, toObject);
```

src/VirtoCommerce.Platform.Core/Common/ReflectionUtility.cs
DmitryGrishinVirtoworks: also I think we can remove "inputType is not null" as "inputType" is the same reference as "type", which
is already guarded by "ArgumentNullException.ThrowIfNull(type)"

DmitryGrishinVirtoworks (approved):

PR #3076, opened 2026-07-13
Fix saving empty dictionary settings

ToxaKZ:
## Description
Fixed an issue where clearing all values of a dictionary setting returned a successful response but did not persist the empty list.
Updated the settings value detection logic so an empty AllowedValues array is treated as a valid dictionary value. Also removed the obsolete workaround that used an empty scalar value to force saving.
### Jira-link:
https://virtocommerce.atlassian.net/browse/VCST-5441
### Artifact URL:

Image tag:
ghcr.io/VirtoCommerce/platform:3.1044.0-pr-3076-3479-vcst-5441-347993a4

### DIFF
--- a/src/VirtoCommerce.Platform.Core/Settings/ObjectSettingEntry.cs
+++ b/src/VirtoCommerce.Platform.Core/Settings/ObjectSettingEntry.cs
@@ -24,7 +24,7 @@ namespace VirtoCommerce.Platform.Core.Settings
             IsHidden = descriptor.IsHidden;
             IsPublic = descriptor.IsPublic;
         }
-        public bool ItHasValues => Value != null || !AllowedValues.IsNullOrEmpty();
+        public bool ItHasValues => Value != null || (IsDictionary ? AllowedValues != null : !AllowedValues.IsNullOrEmpty());
         /// <summary>
         /// Setting may belong to any object in system
         /// </summary>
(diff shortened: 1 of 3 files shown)


OlegoO (approved):

src/VirtoCommerce.Platform.Core/Settings/ObjectSettingEntry.cs
```diff
@@ -24,7 +24,7 @@ public ObjectSettingEntry(SettingDescriptor descriptor)
             IsHidden = descriptor.IsHidden;
             IsPublic = descriptor.IsPublic;
         }
-        public bool ItHasValues => Value != null || !AllowedValues.IsNullOrEmpty();
```
DmitryGrishinVirtoworks, line 27: Did you consider filtering on ItHasValues || existsInDb instead of loosening ItHasValues itself? That would scope the fix to settings that already have a DB row, so untouched dictionary settings with an empty default (e.g. via DeepSaveSettingsAsync) wouldn't start getting persisted as explicit empty overrides. Trade-off: touches more of the method's structure and queries DB for the full candidate set upfront rather than only for non-empty ones, so it's not a clear win either way - not a blocker, just wanted to check if this alternative was on your radar.

DmitryGrishinVirtoworks (approved):

src/VirtoCommerce.Platform.Core/Settings/ObjectSettingEntry.cs
ToxaKZ: Thanks for catching this and pointing out the DeepSaveSettingsAsync scenario. I’ve updated it to skip untouched dictionary settings that have an empty registered default and no persisted DB entry. Existing overrides and explicit empty overrides of non-empty defaults are still saved. This avoids creating redundant empty DB overrides while preserving the original fix.

OlegoO (approved):

PR #3119, opened 2026-09-18
VCST-5508: Fix scope leaks in Func<T> repository and Identity factories

yuskithedeveloper:
### Jira-link:
https://virtocommerce.atlassian.net/browse/VCST-5508
### Artifact URL:

Image tag:
ghcr.io/VirtoCommerce/platform:3.1072.0-pr-3119-9c0a-vcst-5508-func-scope-9c0a099f


yuskithedeveloper:
**Compatibility check for the `protected` legacy constructors** (org-wide GitHub code search over all VirtoCommerce repos + the installed module binaries + a customer module sample)

**Production code: nothing breaks.** No module or customization constructs any of the 17 affected platform classes with `new`, and none derives from them, so prebuilt binaries keep resolving. The legacy `Func<T>` registrations (including the `Func<(IUserSessionsService, IServiceScope)>` tuple used by Customer and ProfileExperienceApi) stay in place.

**Test projects that will fail to compile once the module bumps its platform reference** (they call a legacy constructor with `new`; fix = pass an `IScopedServiceFactory<T>` stub instead of the lambda):

| Repo | File | Call |
|---|---|---|
| vc-module-store | `tests/VirtoCommerce.StoreModule.Tests/StoreNotificationSenderTests.cs` | `new CustomUserManager(..., Func<ISecurityRepository>, ...)` |
| vc-module-sales-rep | `tests/VirtoCommerce.SalesRep.Tests/ComponentTests/Infrastructure/TestServicesConfiguration.cs` | `new UserSearchService(Func<UserManager<ApplicationUser>>, Func<RoleManager<Role>>)` |
| vc-module-ui-migration | copy of the Store test above | `new CustomUserManager(...)` |
| vc-module-DIBS, vc-module-Paypal-ExpressCheckout | payment tests | `new DynamicPropertyService(...)` (v2-era repos, last pushed 2017/2018, out of scope) |

No derived class anywhere calls the obsolete constructors from source, so the VC0016 deprecation does not hit any known repo; it only concerns customizations outside the org.

